PR #114, opened 2020-02-24
Ja/fix client org

monkey-jeff:
This allows you to reference the class directly instead of through a client object

Example:
```
let client = new Product.ShopperProduct({})
```
instead of 
```
let client = new Product.ShopperProduct.Client({});
```

### DIFF
--- a/packages/generator/src/templateHelpers.ts
+++ b/packages/generator/src/templateHelpers.ts
@@ -164,7 +164,7 @@ export const getReturnPayloadType = function(operation: any): string {
       dataTypes.push(
         res.payloads[0].schema.name.value() === "schema"
           ? "Object"
-          : res.payloads[0].schema.name.value()
+          : res.payloads[0].schema.name.value() + "T"
       );
     } else {
       dataTypes.push("void");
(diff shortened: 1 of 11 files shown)


packages/generator/src/templateHelpers.ts
```diff
@@ -164,7 +164,7 @@ export const getReturnPayloadType = function(operation: any): string {
       dataTypes.push(
         res.payloads[0].schema.name.value() === "schema"
           ? "Object"
-          : res.payloads[0].schema.name.value()
+          : res.payloads[0].schema.name.value() + "T"
```
kalagarraj: why + "T" ? any doc-string on what this is?

monkey-jeff: Without it we ran into a conflict with the className and some data types.  
For example if the type was "Customer" and the api was "Customer" they would exist in the same namespace and cause a conflict.  This happened 2 of the mercury APIs.  Doing this ensures Uniqueness.  Once exported within the class namespace we drop the T as it is no longer needed.

bredmond-sf: I think it's a little cleaner to make something like this and call it
const convertToDataTypeName = (name: string) => `${name}T`;

PR #115, opened 2020-02-25
Add two overload functions per function for Response and DTO return types

gkhunger-sf:
Resolves the return type ambiguity by overloading functions. Each function is overloaded by two functions. One returns the DTO when no value for rawResponse is provided. The other returns the DTO or Response object based on the value of rawResponse.

rawResponse is now its own param instead of a field on options object which it was before.

Implementing this solution increased the size of our SDK by ~75KB (Total unpacked SDK size before the change was 1.6MB). I also added extra tsdocs which further increased the size by ~130KB.

### DIFF
--- a/packages/generator/templates/operations.ts.hbs
+++ b/packages/generator/templates/operations.ts.hbs
@@ -2,10 +2,13 @@
   {{#each operations}}
     /**
     * {{description}}
+    * 
+    * If you would like to get a raw Response object use the other {{name}} function.
+    * 
+    * @returns A promise of type {{getReturnPayloadType .}}
     */
     {{name}}(
       options{{#or (is method "get") (is method "delete")}}?{{/or}}: {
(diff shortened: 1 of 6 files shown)


packages/generator/templates/operations.ts.hbs
```diff
@@ -2,10 +2,13 @@
   {{#each operations}}
     /**
     * {{description}}
+    * 
+    * If you would like to get a raw Response object use the other {{name}} function.
+    * 
+    * @returns A promise of type {{getReturnPayloadType .}}
```
kalagarraj: Should this be Response | {{getReturnPayloadType .}}

gkhunger-sf: This overload does not return Response. It's the one below this one that returns it. tsdoc for that one captures that.

PR #117, opened 2020-02-26
Exclude dist from test coverage report for generator

gkhunger-sf:


packages/generator/package.json
```diff
@@ -32,9 +32,10 @@
       "coverage",
       "test",
       "test_integration",
-      "gulpfile.ts",
-      "renderedTemplates",
-      "index.js"
+	  "renderedTemplates",
```
bredmond-sf: I think we're doing this wrong...shouldn't we swap the exclude block for an include 'src'?

gkhunger-sf: makes sense

PR #120, opened 2020-02-27
Update README with better code and screenshots

bredmond-sf:
This updates the sample code and screenshots in the README to be up correct and up to date. It also attempts to address the issue that the screenshots do not appear on npmjs because external relative images won't work there. I'm not sure where to document it, but the images can be inlined on OS X by running:
```
base64 -i images/ResultAutocomplete.jpg | pbcopy
```
to place the base64 encoded image on the clipboard and then pasting it into the field at the bottom of the README.


packages/generator/README.md
```diff
@@ -30,62 +30,92 @@ To use an SDK client, instantiate an object of that client and configure these p
...
 ​
 // Import the SDK
-import { Product, helpers } from 'commerce-sdk';
+import { ClientConfig, helpers, Search } from "commerce-sdk";
 
-helpers.getAuthToken({
+// Create a configuration to use when creating API clients
+const config:ClientConfig = {
+    headers: {
+        connection: "close"
```
unandyala: We don't need this header

PR #122, opened 2020-02-28
Default to closing network connection

gkhunger-sf:
Allow keeping the connection alive by passing keep-alive for Connection header


packages/core/src/base/staticClient.ts
```diff
@@ -57,6 +57,19 @@ async function runFetch(
...
+  // If the user wants to keep the connection alive they can set the Connection
+  // header to 'keep-clive' and we'll respect it. Otherwise, we set it to false.
+  //
+  // explicitly make connection header lowercase because otherwise 
+  // make-fetch-happen makes it lowercase later, creates an array by merging the 
+  // values, and we end up with two values.
+  fetchOptions.headers.connection = fetchOptions.headers.connection
+    ? fetchOptions.headers.connection : fetchOptions.headers.Connection ?
+      fetchOptions.headers.Connection : "close";
+  delete fetchOptions.headers.Connection;
```
bredmond-sf: We should add a couple exported functions to this file to help with setting headers that are pretty much the find/get/set from https://github.com/npm/minipass-fetch/blob/master/lib/headers.js

packages/core/src/base/staticClient.ts
```diff
@@ -57,6 +81,15 @@ async function runFetch(
     ? _.clone(options.client.clientConfig.headers)
     : {};
 
+  // make-fetch-happen sets connection header to keep-alive by default which
+  // keeps node running unless it is explicitly killed.
+  // If the user wants to keep the connection alive they can set the Connection
+  // header to 'keep-alive' and we'll respect it. Otherwise, we set it to "close".
+  const connectionHeader = getHeader("connection", fetchOptions.headers);
```
kalagarraj: What about keeping close as the default and Users can override with keep alive?

gkhunger-sf: That's exactly what we are doing

gkhunger-sf: So we look for a connection header in the headers object provided, if there is already one we leave it alone otherwise we set it to "close"

gkhunger-sf: If it isn't very obvious I can add a little bit more detail in the comment to explain it better.

kalagarraj: Sorry I should have been more clear 
How about keeping it as close in clientConfig as default, and Users can override with keep-alive? The advantage is most of the changes in this method can be removed.

gkhunger-sf: that sounds like a good idea. Let me look into that.

PR #134, opened 2020-03-10
Conditional get requests

kalagarraj:
1. Ensure etag header in the response is cached
2. Ensure cached etag is automatically added as if-none-match header in subsequent requests
3. Ensure last-modified header in the response is cached
4. Ensure cached last-modified is automatically added as if-modified-since header in subsequent requests


packages/core/src/base/staticClient.ts
```diff
@@ -21,15 +21,15 @@ export class ResponseError extends Error {
...
  * @param response the response object containing the dto
- * 
+ *
  * @returns the dto wrapped in a promise
  */
 export async function getObjectFromResponse(
   response: Response
 ): Promise<object> {
-  if (response.ok) {
+  if (response.ok || response.status === 304) {
```
bredmond-sf: Kannan and I discussed a few things that should be captured in comments here:
1. make-fetch-happen will return cached body with status 304 for not modified
2. response.ok is 2xx only
3. Other redirection is on by default for make-fetch-happen, so other 3xx codes can be regarded as errors if they bubble up this far

PR #139, opened 2020-03-11
Update the sample code to work with TS, MJS and CJS

bredmond-sf:


packages/generator/README.md
```diff
@@ -36,11 +36,16 @@ To use an SDK client, instantiate an object of that client and configure these p
...
  */
 ​
-// Import the SDK
-import { ClientConfig, helpers, Search } from "commerce-sdk";
+// Import the SDK in TypeScript or newer versions of Node.js
+import CommerceSdk from "commerce-sdk";
+const { ClientConfig, helpers, Search } = CommerceSdk;
+
+// Older Node.js versions can instead use:
+// const { ClientConfig, helpers, Search } = require("./dist");
```
monkey-jeff: shouldn't be ./dist here

packages/generator/README.md
```diff
@@ -36,11 +36,16 @@ To use an SDK client, instantiate an object of that client and configure these p
  * APIs.
  */
 ​
-// Import the SDK
-import { ClientConfig, helpers, Search } from "commerce-sdk";
+// Import the SDK in TypeScript or newer versions of Node.js
+import CommerceSdk from "commerce-sdk";
+const { ClientConfig, helpers, Search } = CommerceSdk;
```
monkey-jeff: Is this right? I mean `import { ClientConfig, helpers, Search } from "commerce-sdk"` should work fine in ts right?

packages/generator/README.md
bredmond-sf: Good catch! fixed

packages/generator/README.md
bredmond-sf: It does work in ts because we target commonjs. This describes it best:

> With CommonJS, on the other hand, the shape of a module is not known until after the code is evaluated. What this means is, without making significant changes to the ECMAScript language spec, it will not be possible to use Named Imports from a CommonJS module. Instead, developers will be required to use what ES6 Modules call the “default” export.

https://medium.com/the-node-js-collection/an-update-on-es6-modules-in-node-js-42c958b890c

PR #141, opened 2020-03-16
Br/disable cache.w 7007712

bredmond-sf:
This already works and has test coverage. Only doc coverage is required.

### DIFF
--- a/packages/generator/README.md
+++ b/packages/generator/README.md
@@ -36,15 +36,16 @@ To use an SDK client, instantiate an object of that client and configure these p
  * APIs.
  */
 ​
-// Import the SDK in TypeScript or newer versions of Node.js
-import CommerceSdk from "commerce-sdk";
+// Import the SDK in TypeScript
+import * as CommerceSdk from "commerce-sdk";
+// For Javascript, use:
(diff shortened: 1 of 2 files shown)


packages/generator/README.md
```diff
@@ -36,15 +36,16 @@ To use an SDK client, instantiate an object of that client and configure these p
...
+import * as CommerceSdk from "commerce-sdk";
+// For Javascript, use:
+// import as CommerceSdk from "commerce-sdk";
 const { ClientConfig, helpers, Search } = CommerceSdk;
-
 // Older Node.js versions can instead use:
 // const { ClientConfig, helpers, Search } = require("commerce-sdk");
 
 // Create a configuration to use when creating API clients
-// In TypeScript, this can be specified as const config:ClientConfig
```
bethlew: Can you omit "this can be specified as" and instead just have 'const config:ClientConfig

bredmond-sf: That's one the lines that are being removed. You should see the one that's replacing it in green below it.

bethlew: Ah ok!

PR #142, opened 2020-03-17
Add TSDoc for all the classes and methods in Core

gkhunger-sf:


packages/core/src/base/staticClient.ts
```diff
@@ -47,11 +62,12 @@ export async function getObjectFromResponse(
...
- * @param header - target header
- * @param headers - list to search from
- *
- * @returns header from the list if there is a match, the header passed otherwise
+ * 
+ * @param header - Target header
+ * @param headers - List to search from
+ * @returns Header from the list if there is a match, the passed header otherwise
+ * 
+ * @export
```
unandyala: What is @export for? If we are not not using, I suggest to remove it

gkhunger-sf: It is just to state that it is an exported method. It's just extra information. I saw that we were using it in some of our other doc strings and decided to use it. I don't have any strong feelings about using it or not using it. If you feel it is not very useful I can remove it.

unandyala: does not sound very useful

gkhunger-sf: Removed

PR #143, opened 2020-03-17
Render data type for request body

unandyala:
Sample rendered code
```typescript
promotionsSearch(
        options: {
          parameters?: {
            organizationId?: string
            siteId?: string
          },
          headers?: { [key: string]: string },
          body: SearchRequestT
        }
      ): Promise<PromotionSearchResultT>;

 addItemToBasket<T extends boolean>(
        options: {
          parameters?: {
            organizationId?: string
            basketId: string
            siteId?: string
          },
          headers?: { [key: string]: string }, 
          body: Array<ProductItemT>
        },
        rawResponse?: T
      ): Promise<T extends true ? Response : BasketT>;
```

### DIFF
--- a/packages/generator/src/templateHelpers.ts
+++ b/packages/generator/src/templateHelpers.ts
@@ -255,6 +255,37 @@ export const getParameterDataType = function(
   return DEFAULT_DATA_TYPE;
 };
 
+const getPayloadType = function(schema: model.domain.Shape): string {
+  const name = schema.name.value();
+  if (name != null) {
+    if (name === "schema") {
+      return OBJECT_DATA_TYPE;
+    } else {
+      return name + "T";
+    }
+  }
+  return OBJECT_DATA_TYPE;
+};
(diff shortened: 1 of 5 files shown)


packages/generator/src/templateHelpers.ts
```diff
@@ -255,6 +255,37 @@ export const getParameterDataType = function(
   return DEFAULT_DATA_TYPE;
 };
 
+const getPayloadType = function(schema: model.domain.Shape): string {
+  const name = schema.name.value();
+  if (name != null) {
+    if (name === "schema") {
```
kalagarraj: can we avoid nested ifs

```
if (name === null ){
  return OBJECT_DATA_TYPE
}
if(name === 'schema') {
  //do something
} else {
 //do something else
}
```

unandyala: sure

PR #146, opened 2020-03-23
Eviction tests using cache manager

kalagarraj:
New unit tests using cache manager to test cache eviction and updates


packages/core/test/static-client-caching-eviction.test.ts
```diff
@@ -0,0 +1,195 @@
+/*
+ * Copyright (c) 2019, salesforce.com, inc.
```
unandyala: 2020?

kalagarraj: Will take care of it next PR...

PR #148, opened 2020-03-27
Br/redis cache poc

bredmond-sf:
Raising this PR as a good cut point. This is a working redis impl. There will be at least one more PR to include doc updates to README and sample code.


packages/core/src/base/cacheManagerKeyv.ts
```diff
@@ -0,0 +1,230 @@
...
+      (await this.keyv.delete(getContentKey(req)))
+    );
+  }
+
+  /**
+   * Redis will attempt to maintain an active connection which prevents Node.js
+   * from exiting. Call this to gracefully close the connection.
+   */
+  async quit(): Promise<boolean> {
+    return (await this.keyv?.opts?.store?.redis?.quit()) == "OK";
```
kalagarraj: Don't we need strict check for "OK" using "==="?

packages/core/src/base/cacheManagerKeyv.ts
```diff
@@ -0,0 +1,230 @@
...
+        status: 200,
+        size: redisInfo.size
+      })
+    );
+  }
+
+  /**
+   * Takes both a request and its response and adds it to the given cache.
+   *
+   * @param req The request this is a storing a response for
```
kalagarraj: Change to ```@param req The request to be stored```

packages/core/src/base/cacheManagerKeyv.ts
```diff
@@ -0,0 +1,230 @@
...
+      // Update metadata without writing
+      const redisInfo = await this.keyv.get(getMetadataKey(req));
+      // Providing these will bypass content write
+      cacheOpts.integrity = redisInfo.integrity;
+      addCacheHeaders(
+        response.headers,
+        "",
+        ckey,
+        redisInfo.integrity,
+        redisInfo.time
```
kalagarraj: should cache header update the local cache time with current time on 304 response?

i.e, last param could be a new Date()

packages/core/src/base/cacheManagerKeyv.ts
bredmond-sf: We're not storing the request, we're storing the response

packages/core/src/base/cacheManagerKeyv.ts
bredmond-sf: These are the custom headers and it's meant to indicate the time the response was originally cached. This is actually kept the same as the cacache implementation

packages/core/src/base/cacheManagerKeyv.ts
bredmond-sf: Fixed

packages/core/src/base/cacheManagerKeyv.ts
```diff
@@ -0,0 +1,277 @@
...
+ * @param response The response to calculate the TTL for
+ */
+const getTTL = (response: fetch.Response): number => {
+  const responseControl = response.headers
+    .get("cache-control")
+    ?.toLowerCase()
+    .trim()
+    .split(/\s*,\s*/);
+
+  if (responseControl) {
```
kalagarraj: Can we negate this condition and change the flow accordingly? to avoid nested ifs

packages/core/src/base/cacheManagerKeyv.ts
```diff
@@ -0,0 +1,277 @@
...
+import fetch = require("minipass-fetch");
+import Keyv = require("keyv");
+import ssri = require("ssri");
+import url = require("url");
+
+import { ICacheManager } from "./cacheManager";
+
+/**
+ * Calculate a reasonable time to live for the response based on the request and response headers.
+ *
```
monkey-jeff: ?[NIT] Should add an @returns tsdoc string to each of these methods

packages/core/src/base/cacheManagerKeyv.ts
```diff
@@ -0,0 +1,277 @@
...
+ * @param response The response to calculate the TTL for
+ */
+const getTTL = (response: fetch.Response): number => {
```
monkey-jeff: Out of curiosity why this pattern instead of function getTTL?

monkey-jeff: Not camelCase.  Only first character in the abbreviation should be capitalized as our agreed upon camel case styling 
For reference: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

packages/core/src/base/cacheManagerKeyv.ts
bredmond-sf: I wanted to, but I'm having a hard time that. We can't return if there isn't a cache-control header because we need to check for the expires header. But we still need to check the expires header AFTER if there's a cache-control header that doesn't contain anything interesting to us.

packages/core/src/base/cacheManagerKeyv.ts
```diff
@@ -0,0 +1,277 @@
...
+const makeCacheKey = (req: fetch.Request): string => normalizeUrl(req.url);
+
+const getMetadataKey = (req: fetch.Request): string =>
+  `request-cache-metadata:${makeCacheKey(req)}`;
+
+const getContentKey = (req: fetch.Request): string =>
+  `request-cache:${makeCacheKey(req)}`;
+
+const addCacheHeaders = (resHeaders, path, key, hash, time): void => {
+  resHeaders.set("X-Local-Cache", encodeURIComponent(path));
```
monkey-jeff: I don't love this static list of headers.  But can live with it.

packages/core/src/base/cacheManagerKeyv.ts
```diff
@@ -0,0 +1,277 @@
...
+
+    await this.keyv.set(getContentKey(req), body, ttl);
+
+    return Promise.resolve(new fetch.Response(Buffer.from(body), response));
+  }
+
+  /**
+   * Finds the Cache entry whose key is the request, and if found, deletes the
+   * Cache entry and returns a Promise that resolves to true. If no Cache entry
+   * is found, it returns false.
```
monkey-jeff: Missing docstrings

packages/core/src/base/cacheManagerKeyv.ts
```diff
@@ -0,0 +1,230 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+"use strict";
+
+import fetch = require("minipass-fetch");
```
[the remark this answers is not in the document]
bredmond-sf: Fixed. make-fetch-happen is setup in such a way that it only work with the 'require' version. These work without it.

packages/core/package.json
```diff
@@ -6,6 +6,9 @@
   "types": "dist/index.d.ts",
   "scripts": {
     "pretest": "npm run lint",
+    "redis": "docker rm --force redis; docker run -dp 6379:6379 --name redis redis:5-alpine",
```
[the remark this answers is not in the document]
bredmond-sf: Done

packages/core/src/base/cacheManagerKeyv.ts
bredmond-sf: They're mostly interchangeable aside from some very minor scoping differences. I just prefer reading it.

bredmond-sf: Fixed

packages/core/src/base/cacheManagerKeyv.ts
bredmond-sf: Fixed

packages/core/src/base/cacheManagerKeyv.ts
bredmond-sf: Fixed

packages/core/src/base/cacheManagerKeyv.ts
```diff
@@ -0,0 +1,339 @@
...
+   *
+   * @returns true is anything is present to delete, false otherwise
+   */
+  async delete(req: fetch.Request, opts?: any): Promise<boolean> {
+    if (!req) {
+      throw new Error("Valid request object required to delete");
+    }
+    this.stripUncacheableRequestHeaders(req);
+    return (
+      (await this.keyv.delete(getMetadataKey(req))) ||
```
unandyala, line 324: is || not a short circuit evaluation? If both metadata and content are present in cache, don't the if return true as soon as metadata is deleted

bredmond-sf: good catch! fixed

packages/core/src/base/cacheManagerKeyv.ts
```diff
@@ -0,0 +1,339 @@
...
+    if (req.method === "HEAD") {
+      return new fetch.Response(null, {
+        url: req.url,
+        headers: resHeaders,
+        status: 200
+      });
+    }
+
+    // Get the cached response body
+    const body: string = await this.keyv.get(getContentKey(req));
```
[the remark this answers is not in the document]
bredmond-sf: fixed

PR #151, opened 2020-04-03
Ka/ci configure redis

kalagarraj:
- Sets up a remote docker instance for running redis
- npm run test:integration step includes running the redis cache test (Refer to PR-149)


.circleci/config.yml
```diff
@@ -6,9 +6,13 @@ jobs:
     executor:
       name: node/default
       tag: '10.16'
-    
+
     steps:
       - checkout
+      - setup_remote_docker
```
bredmond-sf, line 12: I might move these steps to right before the integration so we won't need to do this at all if the regular unittests fail.

kalagarraj: Done, I was accounting for redis to be ready for the integration tests.

.circleci/config.yml
```diff
@@ -6,32 +6,41 @@ jobs:
...
-    
+
     steps:
       - checkout
       - node/with-cache:
           steps:
             - run: npm install
       - run: npm run build
       - run: npm run test
+      - setup_remote_docker
```
bredmond-sf, line 17: So this isn't actually the approach we want to use:
https://circleci.com/docs/2.0/building-docker-images/#separation-of-environments
this is:
https://circleci.com/docs/2.0/executor-types/#using-multiple-docker-images

kalagarraj: Can u pls clarify? We need redis to run in separate docker and setup remote docker does that.

kalagarraj:
> Do we run tests with both redis and in-memory cache now?

Yes

kalagarraj:
> Do we run tests with both redis and in-memory cache now?

Yes

PR #156, opened 2020-04-09
Remove TTL and fix tests

kalagarraj:
1. Removed ttl method and replaced with cachable or not
2. used keyv method without 3rd optional ttl argument

### DIFF
--- a/packages/core/src/base/cacheManagerKeyv.ts
+++ b/packages/core/src/base/cacheManagerKeyv.ts
@@ -34,9 +34,9 @@ const addCacheHeaders = (resHeaders, path, key, hash, time): void => {
 };
 
 /**
- * Calculate a reasonable time to live for the response based on the response headers.
+ * Determines the response to be cached or not.
  *
- * @param response The response to calculate the TTL for
+ * @param response The response to determine ability for caching
  *
  * @returns boolean to cache the response or not
  */
@@ -47,14 +47,11 @@ const isCachable = (response: fetch.Response): boolean => {
     .trim()
     .split(/\s*,\s*/);
 
-  if (
+  return !(
     responseControl &&
     (responseControl.includes("private") ||
       responseControl.includes("no-store"))
-  ) {
-    return false;
-  }
-  return true;
+  );
 };
 
 /**
--- a/packages/core/src/base/cacheManagerKeyv.ts
+++ b/packages/core/src/base/cacheManagerKeyv.ts
@@ -40,7 +40,7 @@ const addCacheHeaders = (resHeaders, path, key, hash, time): void => {
  *
  * @returns boolean to cache the response or not
  */
-const isCachable = (response: fetch.Response): boolean => {
+const shouldCache = (response: fetch.Response): boolean => {
   const responseControl = response.headers
     .get("cache-control")
     ?.toLowerCase()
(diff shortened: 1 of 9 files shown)


packages/core/src/base/cacheManagerKeyv.ts
```diff
@@ -38,46 +38,23 @@ const addCacheHeaders = (resHeaders, path, key, hash, time): void => {
...
-      return parseInt(sMaxAgeParts[1]);
-    }
-
-    const maxAgeParts = responseControl
-      .find(value => value.startsWith("max-age"))
-      ?.split(/\s*=\s*/);
-    if (maxAgeParts?.length > 1 && parseInt(maxAgeParts[1]) !== NaN) {
-      return parseInt(maxAgeParts[1]);
-    }
+  if (
```
bredmond-sf, line 50: You can drop the if and just return the expression. Otherwise this is good. It would also be good to note here that make-fetch-happen assumes its caches are private, but we're going to assume shared.

packages/core/src/base/cacheManagerKeyv.ts
```diff
@@ -34,50 +34,24 @@ const addCacheHeaders = (resHeaders, path, key, hash, time): void => {
...
+ * Determines the response to be cached or not.
  *
- * @param response The response to calculate the TTL for
+ * @param response The response to determine ability for caching
  *
- * @returns Time to cache the response in seconds, zero for should not be cached
+ * @returns boolean to cache the response or not
  */
-const getTimeToLiveInSeconds = (response: fetch.Response): number => {
+const isCachable = (response: fetch.Response): boolean => {
```
monkey-jeff, line 43: I don't love the name "isCachable" here

Technically it is always cacheable.  we just shouldn't.  What about calling it "shouldCache" or flipping it to "doNotCache"

packages/core/src/base/cacheManagerKeyv.ts
```diff
@@ -167,6 +141,15 @@ export class CacheManagerKeyv implements ICacheManager {
       this.keyv = new Keyv(options.connection, options.keyvOptions);
       this.keyv.on("error", console.error);
     }
+
+    if (this.keyv?.opts?.ttl) {
```
monkey-jeff, line 145: Nice null safe navigation...but if this conditional fails what is the TTL? Should there be a warning?

packages/core/src/base/cacheManagerKeyv.ts
```diff
@@ -264,9 +247,9 @@ export class CacheManagerKeyv implements ICacheManager {
     const size = response?.headers?.get("content-length");
     const metadataKey = getMetadataKey(req);
     const contentKey = getContentKey(req);
-    const ttlInMilliseconds = getTimeToLiveInSeconds(response) * 1000;
+    const cachable = isCachable(response);
 
-    if (ttlInMilliseconds === 0) {
+    if (!cachable) {
```
monkey-jeff, line 252: [NIT] I don't love early returns. But it is fine

packages/core/src/base/cacheManagerKeyv.ts
kalagarraj: Removed log

packages/core/src/base/cacheManagerKeyv.ts
kalagarraj: Done.

packages/core/src/base/cacheManagerKeyv.ts
kalagarraj: Done

PR #160, opened 2020-04-15
Logger for SDK

unandyala:
Logger object is created using loglevel package and exported. Default level is set to WARN

**Usage:**

```javascript
import { sdkLogger } from "@commerce-apps/core"

sdkLogger.setLevel(sdkLogger.levels.INFO);

sdkLogger.warn("Test warning message");
sdkLogger.info("Test info message");
```


unandyala:
> @unandyala can we update the readme with how a user of the SDK may manipulate the logging level too please?

Yes. I am adding the env variable to set the log level. Once I have that, I will document both options

monkey-jeff:
> > @unandyala can we update the readme with how a user of the SDK may manipulate the logging level too please?
> 
> Yes. I am adding the env variable to set the log level. Once I have that, I will document both options

Why are we exposing this as an env variable? I am not sure that that is a good idea.

With what you already have the way anyone can set the logging level is already

```
loglevel.getLogger("COMMERCE_SDK_LOGGER").setLevel("debug")
```
(Perhaps export the logger name as a constant though)

I think that this is an area were developer clarity is paramount and magic env variables often aren't clear.  I would say that if you want to set your log level by env variable most likely they want to do so holistically and not piecemeal.  

I am not hugely opposed to setting the logger with an environmental variable, but I want to make sure what we do doesn't obfuscate logic for the end-user making it more difficult to go from proof of concept to production quality.

One of my pet peeves is when tools are made to make PoC really really easy but then it clunky or non-obvious how to productionalize that code.

unandyala:
Reasoning is user can set the log level without having to touch the code if it is an env variable - change the log level without having to rebuild the code.

As we are exporting the logger, I didn't feel the necessity to expose the key. We can if we think its usefult

monkey-jeff:
> Reasoning is user can set the log level without having to touch the code if it is an env variable - change the log level without having to rebuild the code.
> 
> As we are exporting the logger, I didn't feel the necessity to expose the key. We can if we think its usefult

I just think this is one of those cases where they would use this capability in demo and poc codes but they'd have their own configuration to set logging in a production environment.  Just worried on the transition from demo to prod

unandyala:
Good point. I will add the env variable just for the generator logger and NOT for the sdk logger.

PR #161, opened 2020-04-16
Add example client instantiation to APICLIENTS.md

wjhsf:
- Modified renderer to expose necessary information. Should also make adding generated examples easier in the future.
- Updated [APICLIENTS.md](https://github.com/SalesforceCommerceCloud/commerce-sdk/blob/wjh/update-usage-docs.W-7263996/packages/generator/APICLIENTS.md).
- Added test for `renderOperationList` to meet coverage threshold.

### DIFF
--- a/packages/generator/src/renderer.ts
+++ b/packages/generator/src/renderer.ts
@@ -81,7 +81,7 @@ const dtoTemplate = Handlebars.compile(
   fs.readFileSync(path.join(templateDirectory, "dto.ts.hbs"), "utf8")
 );
 
-const versionTemplate = Handlebars.compile(
+const apiClientsTemplate = Handlebars.compile(
   fs.readFileSync(path.join(templateDirectory, "apiclients.md.hbs"), "utf8")
 );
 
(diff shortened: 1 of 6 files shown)


packages/generator/src/renderer.ts
```diff
@@ -139,6 +163,32 @@ function createHelpers(config: any): string {
...
+  apiFamilyConfig: { [key: string]: RestApi[] }
+): string {
+  const apis = Array.from(apiFamilyMap).map(
+    ([family, apiModels]): ApiClientsInfoT => {
+      // Merge model and config into array of objects
+      return apiModels.map((apiModel: WebApiBaseUnitWithEncodesModel, idx) => {
+        return {
+          family, // Included for ease of access within template
+          model: apiModel.encodes as model.domain.WebApi,
+          config: apiFamilyConfig[family][idx]
```
unandyala, line 183: Will the index match to the same API always in both arrays? Resolving to apiModel is a async call, just make sure the order is retained

wjhsf: The models are generated using `_.map` on the config object, so order is preserved.

https://github.com/SalesforceCommerceCloud/commerce-sdk/blob/c3ec85c59ccdb2172d344cb2b6ea2020e510b6c4/packages/generator/src/parser.ts#L98-L105

PR #162, opened 2020-04-17
updateApis shouldn't fail if it's unable to fetch an asset

gkhunger-sf:
- Display a message to manually download a RAML if exhcnage-downloader couldn't get it
- Increase the code coverage threshold to 85%
- Refactor exchange-downloader tests to use stubs for helper functions

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "private": true,
   "scripts": {
     "postinstall": "lerna bootstrap --hoist",
-    "updateApis": "lerna run updateApis",
+    "updateApis": "lerna run updateApis --stream",
     "build": "lerna run build",
     "test": "lerna run test",
     "test:debug": "lerna run test:debug",
--- a/packages/exchange-connector/gulpfile.ts
+++ b/packages/exchange-connector/gulpfile.ts
@@ -19,11 +19,32 @@ import {
   getSpecificApi,
   groupByCategory
 } from "./src";
+import { sdkLogger } from "../core";
 
 require("dotenv").config();
 
 import config from "../../build-config";
 import { removeRamlLinks } from "./src/exchangeTools";
+
+/**
+ * Remove all the information that could be invalid so that the SDK is not
+ * generated with incorrect information accidentally. Correct information
+ * needs to be filled manually in 'apis/api-config.json' before the SDK can be
+ * generated.
+ *
+ * @param api The API with incorrect information
+ */
+function removeInvalidInformation(api: RestApi) {
+  api.id = null;
+  api.updatedDate = null;
+  api.version = null;
+  api.fatRaml.createdDate = null;
+  api.fatRaml.md5 = null;
+  api.fatRaml.sha1 = null;
+  if (api.fatRaml.externalLink) {
+    delete api.fatRaml.externalLink;
+  }
+}
 
 async function search(): Promise<RestApi[]> {
   const token = await getBearer(
@@ -46,7 +67,7 @@ async function search(): Promise<RestApi[]> {
           if (neededApi) {
             return neededApi;
           } else {
-            api.version = null;
+            removeInvalidInformation(api);
             return api;
           }
         }
@@ -66,7 +87,7 @@ function downloadRamlFromExchange(): Promise<void> {
   return search().then(apis => {
     return downloadRestApis(apis, config.inputDir)
       .then(folder => {
-        console.log(`Setting config.inputDir to '${folder}'`);
+        sdkLogger.warn(`Setting config.inputDir to '${folder}'`);
         config.inputDir = folder;
         return extractFiles(folder);
       })
--- a/packages/exchange-connector/src/exchangeDownloader.ts
+++ b/packages/exchange-connector/src/exchangeDownloader.ts
@@ -12,21 +12,24 @@ import fetch, { Response } from "node-fetch";
 import path from "path";
 
 import { RestApi, FileInfo, Categories } from "./exchangeTypes";
+import { sdkLogger } from "../../core/src";
 
 const DEFAULT_DOWNLOAD_FOLDER = "download";
 const ANYPOINT_BASE_URI = "https://anypoint.mulesoft.com/exchange/api/v2";
+const ANYPOINT_BASE_URI_WITHOUT_VERSION =
+  "https://anypoint.mulesoft.com/exchange";
 
 export function downloadRestApi(
   restApi: RestApi,
   destinationFolder: string = DEFAULT_DOWNLOAD_FOLDER
 ): Promise<void | Response> {
-  return new Promise((resolve, reject) => {
-    if (!restApi.fatRaml) {
-      reject(
-        new Error(
-          `Fat RAML download information for ${restApi.assetId} is missing`
-        )
+  return new Promise(resolve => {
+    if (!restApi.fatRaml.createdDate) {
+      sdkLogger.warn(
+        `Failed to download '${restApi.name}' RAML as Fat RAML download information is missing.`,
+        `Please download it manually from ${ANYPOINT_BASE_URI_WITHOUT_VERSION}/${restApi.assetId} and update the relevant details in apis/api-config.json`
       );
+      return resolve();
     }
 
     ensureDirSync(destinationFolder);
@@ -54,7 +57,7 @@ export function downloadRestApis(
   const promises: Promise<any>[] = [];
 
   restApi.forEach((api: RestApi) => {
-    promises.push(downloadRestApi(api, destinationFolder));
+    promises.push(exports.downloadRestApi(api, destinationFolder));
   });
 
   return Promise.all(promises).then(() => destinationFolder);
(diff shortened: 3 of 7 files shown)


packages/exchange-connector/src/exchangeDownloader.ts
```diff
@@ -12,21 +12,24 @@ import fetch, { Response } from "node-fetch";
...
 
 import { RestApi, FileInfo, Categories } from "./exchangeTypes";
+import { sdkLogger } from "../../core/src";
 
 const DEFAULT_DOWNLOAD_FOLDER = "download";
 const ANYPOINT_BASE_URI = "https://anypoint.mulesoft.com/exchange/api/v2";
+const ANYPOINT_BASE_URI_WITHOUT_VERSION =
+  "https://anypoint.mulesoft.com/exchange";
 
 export function downloadRestApi(
```
bredmond-sf, line 22: Just a suggestion, but this could be simplified a bit...
```
export function async downloadRestApi(
  restApi: RestApi,
  destinationFolder: string = DEFAULT_DOWNLOAD_FOLDER
): Promise<void | Response> {

    if (!restApi.fatRaml?.createdDate) {
      sdkLogger.warn(
        `Failed to download '${restApi.name}' RAML as Fat RAML download information is missing.`,
        `Please download it manually from ${ANYPOINT_BASE_URI_WITHOUT_VERSION}/${restApi.assetId} and update the relevant details in apis/api-config.json`
      );
      return;
    }

    ensureDirSync(destinationFolder);
    const zipFilePath = path.join(destinationFolder, `${restApi.assetId}.zip`);
    const response = await fetch(restApi.fatRaml.externalLink);
    writeFileSync(zipFilePath, response.body);
    return response;
}
```

packages/exchange-connector/src/exchangeDownloader.ts
```diff
@@ -164,24 +170,30 @@ export function getVersionByDeployment(
   restApi: RestApi,
   deployment: RegExp
 ): Promise<string> {
-  return getAsset(accessToken, `${restApi.groupId}/${restApi.assetId}`).then(
-    asset => {
+  return exports
```
bredmond-sf, line 173: same question

packages/exchange-connector/src/exchangeDownloader.ts
```diff
@@ -54,7 +57,7 @@ export function downloadRestApis(
   const promises: Promise<any>[] = [];
 
   restApi.forEach((api: RestApi) => {
-    promises.push(downloadRestApi(api, destinationFolder));
+    promises.push(exports.downloadRestApi(api, destinationFolder));
```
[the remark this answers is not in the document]
gkhunger-sf: to help with stubbing

packages/exchange-connector/gulpfile.ts
```diff
@@ -66,7 +87,7 @@ function downloadRamlFromExchange(): Promise<void> {
   return search().then(apis => {
     return downloadRestApis(apis, config.inputDir)
       .then(folder => {
-        console.log(`Setting config.inputDir to '${folder}'`);
+        sdkLogger.warn(`Setting config.inputDir to '${folder}'`);
```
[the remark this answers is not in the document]
gkhunger-sf: it isn't 😛 
will change to info

packages/exchange-connector/gulpfile.ts
```diff
@@ -19,12 +19,33 @@ import {
   getSpecificApi,
   groupByCategory
 } from "./src";
+import { sdkLogger } from "../core";
```
monkey-jeff, line 22: [ISSUE] Should not have relative dependencies across packages.

packages/exchange-connector/gulpfile.ts
```diff
@@ -19,12 +19,33 @@ import {
...
 
+/**
+ * Remove all the information that could be invalid so that the SDK is not
+ * generated with incorrect information accidentally. Correct information
+ * needs to be filled manually in 'apis/api-config.json' before the SDK can be
+ * generated.
+ *
+ * @param api The API with incorrect information
+ */
+function removeInvalidInformation(api: RestApi) {
```
monkey-jeff, line 37: The method should not be in the gulpfile.  Gulpfiles should be as thin as possible.

packages/exchange-connector/src/exchangeDownloader.ts
```diff
@@ -12,21 +12,24 @@ import fetch, { Response } from "node-fetch";
 import path from "path";
 
 import { RestApi, FileInfo, Categories } from "./exchangeTypes";
+import { sdkLogger } from "../../core/src";
```
monkey-jeff, line 15: Same as above.  Don't use relative imports across packages.

packages/exchange-connector/src/exchangeDownloader.ts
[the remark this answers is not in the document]
monkey-jeff: I have to say...I don't love this but will accept it if we don't have a better way.

packages/exchange-connector/src/exchangeDownloader.ts
monkey-jeff: I like this proposal

packages/exchange-connector/src/exchangeDownloader.ts
monkey-jeff: I like this even less here.

packages/exchange-connector/test/exchangeDownloader.test.ts
```diff
@@ -67,84 +68,76 @@ describe("Test downloadRestApi", () => {
...
-      "Fat RAML download information for test-api is missing"
-    );
+    return expect(downloadRestApi(api)).to.eventually.be.undefined;
   });
 });
 
-describe("Test downloadRestApis", () => {
-  afterEach(nock.cleanAll);
+describe("downloadRestApis", () => {
+  beforeEach(sinon.restore);
```
monkey-jeff, line 96: Why did you switch from nock to sinon here?

packages/exchange-connector/test/exchangeDownloader.test.ts
```diff
@@ -153,30 +146,28 @@ describe("Test searchExchange", () => {
...
 
     return searchExchange("AUTH_TOKEN", "searchString").then(res => {
-      expect(res).to.be.deep.equal(searchAssetApiResultObject);
+      expect(res).to.deep.equal(searchAssetApiResultObject);
     });
   });
 });
 
-describe("Test getSpecificApi", () => {
-  afterEach(nock.cleanAll);
```
monkey-jeff, line 162: Same question

packages/generator/src/parser.ts
```diff
@@ -15,6 +15,7 @@ import amf from "amf-client-js";
 import path from "path";
 import _ from "lodash";
 import { RestApi } from "@commerce-apps/exchange-connector";
+import { sdkLogger } from "../../core/src";
```
monkey-jeff, line 18: Same as before.  Do not have cross package relative dependencies.

packages/generator/src/parser.ts
```diff
@@ -97,6 +98,10 @@ export function processApiFamily(
   const promises = [];
   const ramlFileFromFamily = apiFamilyConfig[apiFamily];
   _.map(ramlFileFromFamily, (apiMeta: RestApi) => {
+    if(!apiMeta.fatRaml.createdDate) {
```
monkey-jeff, line 101: Why are we only checking for createdDate?

package.json
```diff
@@ -3,7 +3,7 @@
   "private": true,
   "scripts": {
     "postinstall": "lerna bootstrap --hoist",
-    "updateApis": "lerna run updateApis",
+    "updateApis": "lerna run updateApis --stream",
```
monkey-jeff, line 6: Why this change here?

packages/exchange-connector/src/exchangeDownloader.ts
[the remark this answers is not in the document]
gkhunger-sf: I know it's not ideal but it's either this or moving the helper functions to a separate file.

packages/exchange-connector/test/exchangeDownloader.test.ts
gkhunger-sf: To stub the helper functions, sinon was only mocking the calls made by the helper functions. I believe the tests shouldn't be invoking the helper functions unless they are the ones under test.

packages/generator/src/parser.ts
gkhunger-sf: createdDate is one of the fields that we reset when we are unable to get an asset from exchange. Since it is one of the mandatory fields, we shouldn't go forward if this is not set.

package.json
gkhunger-sf: lerna doesn't show logs from the child processes by default, it only shows the end result

packages/exchange-connector/src/exchangeDownloader.ts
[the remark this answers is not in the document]
unandyala: I don't think we want this approach, I documented the details in the renderer tests
```javascript
/**
 * Tests all the functions that are invoked while rendering templates.
 *
 * Note: Ideally we want a independent test for each function. As of today (01/30/2020) this is not possible with Sinon when the functions are in the same module.
 * Sinon works by overwriting a reference to a function on an object, making it point at something else. It can't replace references to the same function in other contexts.
 *
 * Example: We can't mock renderApiFamily function while testing renderTemplates function. To make it work renderTemplates should invoke
 * renderApiFamily as "module.exports.renderApiFamily(..)" which is not desired.
 */
```

PR #163, opened 2020-04-17
generator logger

unandyala:
- Logger for generator
- Added unit test for both SDK logger and generator logger
- Updated README.md and GENERATOR.md

### DIFF
--- a/packages/generator/README.md
+++ b/packages/generator/README.md
@@ -151,9 +151,10 @@ Default log level of the SDK is WARN (warning). SDK uses [loglevel](https://www.
 
 To change the loglevel, import sdkLogger and set the level on it.
 ```javascript
-import { sdkLogger } from "@commerce-apps/core"
+import * as CommerceSdk from "commerce-sdk";
+import * as log from "loglevel";
 
-sdkLogger.setLevel(sdkLogger.levels.INFO);
+log.getLogger(CommerceSdk.COMMERCE_SDK_LOGGER_KEY).setLevel(log.levels.INFO);
 ```
 
 ## Additional Documentation 
--- a/packages/generator/README.md
+++ b/packages/generator/README.md
@@ -149,7 +149,7 @@ When the specified amount of memory is reached, Redis can be configured to apply
 ## Logging
 Default log level of the SDK is WARN (warning). SDK uses [loglevel](https://www.npmjs.com/package/loglevel) npm package. All the log levels supported by [loglevel](https://www.npmjs.com/package/loglevel) package are supported in SDK.
 
-To change the loglevel, import sdkLogger and set the level on it.
+To change the loglevel, set the desired level on the SDK logger.
 ```javascript
 import * as CommerceSdk from "commerce-sdk";
 import * as log from "loglevel";
--- a/packages/generator/README.md
+++ b/packages/generator/README.md
@@ -146,6 +146,15 @@ const config = {
 
 When the specified amount of memory is reached, Redis can be configured to apply an eviction policy. Refer to [this article](https://redis.io/topics/lru-cache/) to setup Redis as an LRU cache and to learn more about supported eviction policies. 
 
+## Logging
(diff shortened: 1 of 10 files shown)


packages/generator/README.md
```diff
@@ -146,6 +146,15 @@ const config = {
 
 When the specified amount of memory is reached, Redis can be configured to apply an eviction policy. Refer to [this article](https://redis.io/topics/lru-cache/) to setup Redis as an LRU cache and to learn more about supported eviction policies. 
 
+## Logging
+Default log level of the SDK is WARN (warning). SDK uses [loglevel](https://www.npmjs.com/package/loglevel) npm package. All the log levels supported by [loglevel](https://www.npmjs.com/package/loglevel) package are supported in SDK.
+
+To change the loglevel, import sdkLogger and set the level on it.
+```javascript
+import { sdkLogger } from "@commerce-apps/core"
```
monkey-jeff, line 154: [ISSUE] We don't have customers using core for anything else, we shouldn't here either. Suggest changing to

```javascript
loglevel.getLogger("COMMERCER-SDK").setLevel(INFO);
```

Alternatively, you can export the SDK logger from commerce-sdk.  But we shouldn't confuse the users by adding. a new package that they need to import.

unandyala: the logger key, I have is "COMMERCE_SDK_LOGGER". Do we want to change to "COMMERCER-SDK"?  I prefer having LOGGER in the key though

wjhsf: The key is only ever used within the context of the logger. Including LOGGER in the key just seems redundant to me.

unandyala: loglevel.getLogger function returns a new logger if the name does not match. If there is a typo in the logger key, a new logger is returned, no error is thrown.

I think I will export the key and use it

packages/generator/README.md
```diff
@@ -146,6 +146,16 @@ const config = {
...
 
+## Logging
+Default log level of the SDK is WARN (warning). SDK uses [loglevel](https://www.npmjs.com/package/loglevel) npm package. All the log levels supported by [loglevel](https://www.npmjs.com/package/loglevel) package are supported in SDK.
+
+To change the loglevel, set the desired level on the SDK logger.
+```javascript
+import * as CommerceSdk from "commerce-sdk";
+import * as log from "loglevel";
+
+log.getLogger(CommerceSdk.COMMERCE_SDK_LOGGER_KEY).setLevel(log.levels.INFO);
```
monkey-jeff, line 157: Hmm, I don't like having to take two components to do something like this.  I would suggest only needing 'commerce-sdk' or only needing 'loglevel' to modify the level.  Otherwise, it isn't crystal clear to the end-user.  

I don't really care which way you go but I suggest we have it be something like
```
import { Logger } from "commerce-sdk"
Logger.setLevel(Logger.INFO);
```
OR something
```
import * as log from "loglevel"
log.getLogger("COMMERCE-SDK").setLevel(log.levels.INFO);
```

What we should avoid is showing you need two distinct components to set something like the logging level.

(Note: I somewhat prefer the first option, but could be convinced either way)

unandyala: I too prefer the first one. It matches with the way that we use to use the logger

packages/generator/README.md
```diff
@@ -151,10 +151,9 @@ Default log level of the SDK is WARN (warning). SDK uses [loglevel](https://www.
 
 To change the loglevel, set the desired level on the SDK logger.
 ```javascript
-import * as CommerceSdk from "commerce-sdk";
-import * as log from "loglevel";
+import { sdkLogger } from "commerce-sdk";
 
-log.getLogger(CommerceSdk.COMMERCE_SDK_LOGGER_KEY).setLevel(log.levels.INFO);
+sdkLogger.setLevel(sdkLogger.levels.INFO);
```
monkey-jeff, line 156: Have you tested that this works? (Is levels exported from here?)

unandyala: yes, "log" is the root logger. All the features of root logger are available in named logger as well. updated my unit tests to use the same way.

PR #164, opened 2020-04-21
Update the gulpfiles, core and generator packages to use EC in raml-toolkit

gkhunger-sf:
The build is failing because raml-tolkit with exhcange-connector hasn't been published yet


packages/generator/gulpfiles/gulpfile-exchange-connector.ts
```diff
@@ -0,0 +1,96 @@
+/*
```
unandyala, line 1: We decided to keep all the file names in lower camelCase. There was a PR before to rename all the existing files. Since these files are tasks I would suggest rename something like **apiDownloadTask**

packages/generator/tsconfig.json
```diff
@@ -6,7 +6,7 @@
     },
     "exclude": [
         "node_modules",
-        "gulpfile.ts",
+        "gulpfiles",
```
unandyala, line 9: Is this still valid after the rename?

gkhunger-sf: yeah, because this is the name of the directory and it is still the same.

PR #168, opened 2020-04-23
Br/add info level logging.w 7365183

bredmond-sf:
Add info level logging

As an app developer I want info level logging from the SDK so that I can debug issues with my app.

Acceptance Criteria
Log info when API request made
API end point
Log info for result
Response code
Log if the response was OK (per make fetch happen).
Also log OK if response code is 304
Status text

Test plan
unit test

Example:
```
Request: POST https://0dnz6oep.api.commercecloud.salesforce.com/customer/shopper-customers/v1/organizations/f_ecom_bblx_stg/customers/actions/login?clientId=b6945b0e-fa1b-4232-a562-a5b34462aa02&siteId=RefArch
Received successful response: 200 OK
Request: GET https://0dnz6oep.api.commercecloud.salesforce.com/search/shopper-search/v1//organizations/f_ecom_bblx_stg/product-search?q=dress&siteId=RefArch
Received successful response: 200 OK
```


packages/core/src/base/staticClient.ts
```diff
@@ -78,6 +79,25 @@ export function getHeader(
   return header;
 }
 
+/**
+ * Format the request being made for logging.
+ *
+ * @param resource The resource being requested
+ * @param fetchOptions The options to the fetch call
+ */
+export const formatFetchForInfoLog =
```
monkey-jeff, line 88: [NIT] really minor...but should this be `formatRequestForInfoLog` instead

packages/core/src/base/staticClient.ts
```diff
@@ -78,6 +79,25 @@ export function getHeader(
...
+  (resource:string, fetchOptions:RequestInit): string => `Request: ${fetchOptions.method.toUpperCase()} ${resource}`;
+
+/**
+ * Format the response received for logging.
+ *
+ * @param response The response received
+ */
+export const formatResponseForInfoLog = (response:Response): string => {
+  const successString = (response.ok || response.status === 304) ? "successful" : "unsuccessful";
+  return `Received ${successString} response: ${response.status} ${response.statusText}`;
```
monkey-jeff, line 98: Consider making it more uniform with the previous statement

```suggestion
  return `Response: ${successString} - ${response.status} ${response.statusText}`;
```

Or something like that

packages/core/src/base/staticClient.ts
```diff
@@ -139,8 +159,12 @@ async function runFetch(
     fetchOptions.cacheManager = options.client.clientConfig.cacheManager;
   }
 
+  sdkLogger.info(formatFetchForInfoLog(resource, fetchOptions));
+
   const response = await fetch(resource, fetchOptions);
 
+  sdkLogger.info(formatResponseForInfoLog(response));
```
monkey-jeff, line 166: Do we have a test that ensures that this is called on a failure?

packages/core/src/base/staticClient.ts
bredmond-sf: I had thought that, but it felt misleading since there is a Request object and that's not what this is. The input is the arguments to the fetch call including the fetch options. It ends up being more fetch specific than Request specific.

packages/core/src/base/staticClient.ts
bredmond-sf: We have tests that ensure a return from this function on failures like 404's and go down this path.

packages/core/src/base/staticClient.ts
bredmond-sf: I just copied the format from the AWS SDK 😹But I like yours, updated.

PR #170, opened 2020-04-24
Add User-Agent header to track SDK usage

wjhsf:
* The static client test file was quite large, so I split it and moved the files into a subdirectory.
* Cleaned up the tests a bit, particularly the headers tests.
* Added `User-Agent` header to all requests. The format is `commerce-sdk@<version>;`, e.g. `commerce-sdk@1.4.0-beta.0;`.


packages/core/src/base/staticClient.ts
```diff
@@ -13,7 +13,10 @@ export { DefaultCache, Response };
 import { Resource } from "./resource";
 import { BaseClient } from "./client";
 import { sdkLogger } from "./sdkLogger";
+import pkg from "../../package.json";
 
+// Version is from @commerce-apps/core, but it will always match commerce-sdk
```
kalagarraj, line 18: +1 for calling this out..

kalagarraj: pls add or update a test in integration to verify this new header matching the commerce-sdk version

wjhsf: I have added a test.

wjhsf:
I originally used `import pkg from "../../package.json")`. This caused TypeScript to complain, and fixing that complaint required a complex solution with multiple `tsconfig.json` files.

I changed the import to `const pkg = require("../../package.json")` and added an `eslint-ignore` directive. This solution doesn't require any other changes, but I'm not sure whether it will cause issues downstream. Is `@typescript-eslint/no-var-requires` a best-practices kind of rule or things-will-break kind of rule?

monkey-jeff:
> I originally used `import pkg from "../../package.json")`. This caused TypeScript to complain, and fixing that complaint required a complex solution with multiple `tsconfig.json` files.
> 
> I changed the import to `const pkg = require("../../package.json")` and added an `eslint-ignore` directive. This solution doesn't require any other changes, but I'm not sure whether it will cause issues downstream. Is `@typescript-eslint/no-var-requires` a best-practices kind of rule or things-will-break kind of rule?

This is the proper solution here.

packages/core/src/base/staticClient.ts
```diff
@@ -152,6 +175,12 @@ async function runFetch(
     _.merge(fetchOptions.headers, options.headers);
   }
 
+  // Delete all user-specified user agents to prevent an override
+  // (Multiple can be specified by using different casing)
+  stripHeaders("user-agent", fetchOptions.headers);
+  // Specify user agent using lower case in order to override make-fetch-happen
+  fetchOptions.headers["user-agent"] = USER_AGENT;
```
monkey-jeff, line 182: Why do we have to strip it if we are just setting it here anyway?  Seems needless.

kalagarraj: User agent is used to determine then response payload in some retail web architectures. I would just add the commerce sdk user agent without stripping others.

wjhsf: This user agent is only sent to our own APIs. Our APIs don't have any conditional behavior based on user agent.

PR #174, opened 2020-05-05
Ja/retry settings.w 7391789

monkey-jeff:
Added the ability to set retry settings to calls.

i.e.

```
product.getProduct({
  retrySettings: {
    retries: 2
  }
});
```

or in the client
```
let client = new Product({
  retrySettings: {
    retries: 5
  }
});
```


wjhsf:
This PR doesn't follow the new [documentation guidelines](https://salesforce.quip.com/2i68A7nnYUXa#HddACAk3gSn). :-1:

packages/core/src/base/staticClient.ts
```diff
@@ -151,42 +124,40 @@ async function runFetch(
...
+  // Headers are treated separately to be able to move them into their own object.
+  const headers = new Headers(_.merge({}, options.client.clientConfig.headers));
 
-  // To disable response caching, set cacheManager to null
-  if (options.client.clientConfig.cacheManager) {
-    fetchOptions.cacheManager = options.client.clientConfig.cacheManager;
+  for (const [header, value] of Object.entries(options.headers || {})) {
+    headers.set(header, value);
   }
+  finalOptions["headers"] = headers;
```
wjhsf, line 153: Why is this `["headers"]` and not `.headers`?

packages/core/src/base/staticClient.ts
```diff
@@ -109,17 +88,10 @@ export const logResponse = (response: Response): void => {
...
-      `${msg}\nResponse Headers: ${JSON.stringify(
-        response.headers.raw(),
-        null,
-        2
-      )}`
-    );
-  } else if (sdkLogger.getLevel() <= sdkLogger.levels.INFO) {
-    sdkLogger.info(msg);
-  }
+  sdkLogger.info(msg);
```
unandyala, line 91: This is fine. The reason I had it as one message (by concatenating) is - we/sdk users don't have to try to link the messages (response with its headers in this case) when there are messages that are logged asynchronously

packages/core/src/base/staticClient.ts
monkey-jeff: Typescript didn't know what the type was in the IDE as it was merged from lodash.  This is so it won't complain about missing .headers during the assignment.

packages/core/src/base/staticClient.ts
monkey-jeff: Totally agree...but we should do this with a request id or a log id that traces a path throw the system.

packages/core/test/static-client.test.ts
```diff
@@ -35,297 +34,299 @@ const CONNECTION_KEEP_ALIVE = { connection: "keep-alive" };
...
+    expect(nock.isDone()).to.be.true;
   });
 
-  it("deletes resource with id in query param and returns 200", () => {
-    const scope = nock("https://somewhere")
-      .delete("/over/the")
-      .query({ id: "rainbow" })
+  it("deletes resource with id in query param and returns 200", async () => {
+    nock("https://delete.test")
+      .delete("/delete")
```
dbarbato-sf, line 95: This is not a unique endpoint, see line 58.

packages/core/package.json
```diff
@@ -124,12 +125,13 @@
     "keyv": "^4.0.0",
     "lodash": "~4.17.15",
     "loglevel": "^1.6.8",
-    "make-fetch-happen": "~7.1.0",
+    "make-fetch-happen": "^8.0.4",
     "minipass-fetch": "^1.2.1",
     "qs": "~6.9.1",
-    "snyk": "^1.316.1",
+    "retry": "^0.12.0",
```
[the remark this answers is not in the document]
monkey-jeff: https://github.com/tim-kos/node-retry/blob/master/License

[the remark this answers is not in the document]
monkey-jeff: https://gus.lightning.force.com/lightning/r/ADM_Third_Party_Software__c/a0qB0000000KPpLIAW/view

packages/core/src/base/client.ts
```diff
@@ -30,15 +36,26 @@ export class ClientConfig {
   public baseUri?: string;
   public cacheManager?: ICacheManager;
   public headers?: { [key: string]: string };
+  public appendHeaders?: { [key: string]: string };
```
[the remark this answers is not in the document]
monkey-jeff: There is right below where I am setting it in the default.

packages/generator/templates/operations.ts.hbs
```diff
@@ -94,6 +97,7 @@
         path: "{{../path}}",
         pathParameters: pathParameters,
         queryParameters: queryParameters,
+        retrySettings: (options || {}).retrySettings,
```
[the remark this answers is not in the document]
monkey-jeff: Just following the same patterns as below.  We can refactor if we want to.  This is how you do null safe navigation in javascript.  I would want to make sure we have been integration testing before we refactor the way we do this.

packages/core/src/base/client.ts
```diff
@@ -30,15 +36,26 @@ export class ClientConfig {
...
   cacheManager: new DefaultCache(
     tmp.dirSync({ prefix: "cache-", unsafeCleanup: true }).name
   ),
-  headers: {},
+  headers: {
+    "content-type": "application/json",
+    connection: "close"
+  },
+  // These are headers that are always Added to all calls
+  // The can still technically be disabled by overwriting this in the config.
```
kalagarraj, line 54: Change to `This can still...`

packages/core/src/base/client.ts
```diff
@@ -30,15 +36,26 @@ export class ClientConfig {
...
   // Enables cacache for local caching in temp dir by default, unsafeCleanup == rm -rf on exit
   cacheManager: new DefaultCache(
     tmp.dirSync({ prefix: "cache-", unsafeCleanup: true }).name
   ),
-  headers: {},
+  headers: {
+    "content-type": "application/json",
+    connection: "close"
+  },
+  // These are headers that are always Added to all calls
```
kalagarraj, line 53: `Added` is capitalized..change to lower case

packages/core/src/base/client.ts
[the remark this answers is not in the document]
kalagarraj: It is definitely a non standard attribute when compared to the rest of the client config ones. I think something like the following would be apt. Thoughts?
`
public customAttributes?: { 
  appendHeaders?: { [key: string]: string };
};
`

packages/core/src/base/client.ts
```diff
@@ -30,15 +36,26 @@ export class ClientConfig {
...
+  // The can still technically be disabled by overwriting this in the config.
+  // This is something I am okay with as it is an explicit action. And think we should avoid magic that prevent explicit actions.
```
bredmond-sf, line 55:
```suggestion
  // This is something I am okay with as it is an explicit action. And think we should avoid magic that prevents explicit actions.
```

PR #182, opened 2020-05-13
Refactor circle ci config and add a job for node 14

gkhunger-sf:


.circleci/config.yml
```diff
@@ -1,79 +1,105 @@
...
-  build-12:
-    environment:
-      COMMERCE_SDK_INPUT_DIR: ../test/raml/valid/
-    working_directory: ~/build_only
+# EXECUTORS
+executors:
+  node-10:
+    <<: *defaults
+    docker:
+      - image: circleci/node:10.16.3
```
bredmond-sf, line 28: Maybe a good time to upgrade? https://nodejs.org/en/about/releases/

.circleci/config.yml
```diff
@@ -1,79 +1,105 @@
...
+      - image: redis:5-alpine
+  node-12:
+    <<: *defaults
     docker:
       - image: circleci/node:12.16.1
       - image: redis:5-alpine
+  node-14:
+    <<: *defaults
+    docker:
+      - image: circleci/node:14
```
bredmond-sf, line 38: We should be consistent about what version we use and should specify the full version here.

.circleci/config.yml
```diff
@@ -1,79 +1,105 @@
-jobs:
-  build:
-    environment:
-      COMMERCE_SDK_INPUT_DIR: ../test/raml/valid/
-    working_directory: ~/build_only
-    docker:
-      - image: circleci/node:10.16.3
-      - image: redis:5-alpine
+# DEFAULTS
```
bredmond-sf, line 1: Nicely done

gkhunger-sf: Thanks 🙂

PR #186, opened 2020-05-14
Add eslint tsdoc with fixes - w 7198851

bredmond-sf:
As a Commerce Cloud app developer I want to have detailed method information in my TSDOC for the generated methods

* Docstrings should have a description, parameters (including type where applicable, even when “self-explanatory”), and return
* TSDOC should take description from RAML file wherever applicable.
* SDK specific parameters can be hard coded.
* TSDOC should output associated method
* Utilize eslint TSDOC plugin

Before:
Screen Shot 2020-05-14 at 10 47 37 AM

After (including typescript sig for reference):
Screen Shot 2020-05-13 at 12 49 15 AM
Screen Shot 2020-05-13 at 12 49 30 AM


packages/generator/test/template-helpers-format-for-tsdoc.test.ts
```diff
@@ -0,0 +1,83 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+"use strict";
+
+import { expect } from "chai";
+
+import { formatForTsDoc } from "../src/templateHelpers";
+
+describe("Test formatForTsDoc template help function", () => {
```
wjhsf, line 13: The first half of these tests are `.to.eq()` and the second half are `.to.be.eq()` :|

bredmond-sf: good catch

packages/generator/src/renderer.ts
```diff
@@ -166,9 +182,9 @@ function createHelpers(config: any): string {
 
 /**
  * Render the API Clients markdown file using the Handlebars template
- * @param {Map<string, WebApiBaseUnit[]>} apiFamilyMap
- * @param {Object.<string, RestApi[]} apiFamilyConfig
- * @returns {string} The rendered template
+ * @param apiFamilyMap -
+ * @param apiFamilyConfig -
```
gkhunger-sf, line 186: Missing description

bredmond-sf: Yea, no one responded to me when I asked about this one in our chat :) I'm happy to add one, I just don't know what it is.

monkey-jeff: What do you think the description should be?

wjhsf: That's why you git blame and @ping. 😛 I was going to reply, but forgot!

```
* @param apiFamilyMap - Collection of API names and the AMF models associated with each API
* @param apiFamilyConfig - The API family config
```

bredmond-sf: done

PR #194, opened 2020-05-20
Updated circleci config to override env vars for publish

monkey-jeff:


.circleci/config.yml
```diff
@@ -60,6 +60,9 @@ jobs:
 
   publish:
     executor: node-10
+    working_directory: ~/publish
+    environment:
+      COMMERCE_SDK_INPUT_DIR: ../test/raml/valid/
```
gkhunger-sf, line 65: You forgot to remove the env var.

PR #195, opened 2020-05-20
Use custom config for publish job

gkhunger-sf:
Dry run of npm publish with these settings worked as expected - https://app.circleci.com/pipelines/github/SalesforceCommerceCloud/commerce-sdk/1282/workflows/05fe13d5-f2d2-433b-8183-356f567bdf17/jobs/2115

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,10 @@ jobs:
       - test
 
   publish:
-    executor: node-10
+    executor:
+      name: node/default
+      tag: '10.20.1'
+    working_directory: ~/publish
     steps:
       - build
       - run:
@@ -83,7 +86,7 @@ workflows:
       - build-node-14:
           filters:
             tags:
-              only: /^v.*/
+              ignore: /^v.*/
 
   build-test-deploy:
     jobs:


.circleci/config.yml
```diff
@@ -59,7 +59,10 @@ jobs:
       - test
 
   publish:
-    executor: node-10
+    executor:
+      name: node/default
```
dbarbato-sf, line 63: Just curious, what does node/default refer to?

.circleci/config.yml
```diff
@@ -59,7 +59,10 @@ jobs:
...
+      tag: '10.20.1'
```
dbarbato-sf, line 64: Also curious what this is referring to.

.circleci/config.yml
gkhunger-sf: It's the pre-built docker image we want circle to use

.circleci/config.yml
gkhunger-sf: circle has image tags for each node version. Here's the complete list of pre-built docker images and tags - https://circleci.com/docs/2.0/docker-image-tags.json. With `name` and `tag`, we are telling circle to use the docker image that has the node version we want to use.

PR #198, opened 2020-05-26
Fix support for Redis cluster cache manager.

wjhsf:
[@keyv/redis](https://www.npmjs.com/package/@keyv/redis) doesn't support Redis clusters, but there is an [open PR](https://github.com/lukechilds/keyv-redis/pull/37) to add support. In the meantime, we need to implement a workaround.

We log requests as JSON, but ioredis clusters contain circular references. To get around that, I changed the logging to replace cluster instances with a static string, `<Redis Cluster>`. There are probably better ways to handle that. ¯\\\_(ツ)\_/¯

Added tests to verify that passing different options to `CacheManagerRedis` correctly connects to either a single instance or a cluster.


packages/core/test/cacheManagerRedis.test.ts
```diff
@@ -46,15 +43,36 @@ describe("quit tests", () => {
...
   it("returns true when called with working redis", async () => {
     cacheManager.keyv = {
-      opts: { store: { redis: { quit: async () => "OK" } } }
+      opts: { store: { redis: { quit: async (): Promise<string> => "OK" } } }
     };
     return expect(cacheManager.quit()).to.eventually.be.true;
   });
 });
+
+describe("Redis connections", () => {
```
kalagarraj, line 61: Add test creating cache manager without connection and clusterConfig..

packages/core/src/base/cacheManagerRedis.ts
```diff
@@ -25,9 +25,21 @@ export class CacheManagerRedis extends CacheManagerKeyv {
     clusterConfig?: {};
   }) {
     if (options?.clusterConfig) {
-      options.keyvStore = new KeyvRedis(
-        new Redis.Cluster(options.clusterConfig)
-      );
+      // TODO: Remove workaround when cluster support is added (hopefully soon):
```
unandyala, line 28: Do we have a backlog story for this?

wjhsf: Yes.

packages/core/src/base/staticClient.ts
```diff
@@ -70,7 +71,8 @@ export function logFetch(resource: string, fetchOptions: RequestInit): void {
   sdkLogger.debug(
     `Fetch Options: ${JSON.stringify(
       fetchOptions,
-      null,
+      // Redis clusters have circular references and can't be converted to JSON
```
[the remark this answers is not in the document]
wjhsf: This was the cleanest way I could successfully log the object with minimal loss of information. If we don't want to log everything, what should we log instead?

packages/core/test/cacheManagerRedis.test.ts
wjhsf: Added tests for CacheManagerKeyv as the parent class.

packages/core/src/base/staticClient.ts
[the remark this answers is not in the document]
jbegleiter-sf: What about https://redis.io/commands/cluster-info ?

packages/core/test/cacheManagerKeyv.constructor.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    const cacheManager = new CacheManagerKeyv({ keyvStore });
+    expect(cacheManager.keyv.opts.store).to.equal(keyvStore);
+  });
+
+  it("accepts a database connection string to create the keyv store", () => {
+    const cacheManager = new CacheManagerKeyv({
+      connection: "redis://localhost"
+    });
+    const { redis } = cacheManager.keyv.opts.store;
+    redis.quit();
```
jbegleiter-sf, line 29: What's up with quit?

wjhsf: Creating the cache manager opens a connection. We're not using it in this test, so we just close it right away so it doesn't block the script from exiting.

packages/core/src/base/staticClient.ts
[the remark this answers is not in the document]
wjhsf: I haven't looked into how to get that info using ioredis, but I assume it would be an async method. That wouldn't play nice with JSON.stringify.

PR #199, opened 2020-05-28
Br/refactor parser

bredmond-sf:
This will not work until the other PR is merged for raml-toolkit, but I wanted to get this open. This will need the raml-toolkit published first and then the package.json updated here before this can be merged, but that is the only change to come.


packages/generator/src/renderer.ts
```diff
@@ -282,6 +299,38 @@ function renderApiFamily(
...
+      Please ensure that '${apiMeta.name}' RAML and its dependencies are present in 'apis/', and all the required information is present in 'apis/api-config.json'.`);
+    }
+    promises.push(
+      parseRamlFile(
+        path.join(inputDir, apiMeta.assetId, apiMeta.fatRaml.mainFile)
+      )
+    );
+  });
+
+  return promises;
```
unandyala, line 331: does it make sense to wait for all the promises to resolve and then return the array of apis instead of promises

packages/generator/test/renderer.test.ts
```diff
@@ -98,13 +97,23 @@ describe("Rendering Tests", () => {
...
+          renderer.processApiFamily(apiGroup, apiConfig, apiInputDir)
         );
       })
     );
     const list = renderer.renderOperationList(allApis);
     expect(list).to.be.a("string");
   });
+
+  it("Throws error when id is missing", async () => {
+    // expect().to.be.rejected was not working here
```
unandyala, line 109: just curious..is this because the function returns  array of unresolved promises?

wjhsf: It's not working because `chai-as-promised` is loaded as a plugin.

packages/generator/src/renderer.ts
wjhsf: My current work also touches this function, and I have already fixed it on my branch. You can leave this for now.

packages/generator/test/template-helpers-get-base-uri.test.ts
```diff
@@ -7,27 +7,25 @@
 "use strict";
 
 import { expect } from "chai";
-import * as lib from "webapi-parser";
-const wap = lib.WebApiParser;
-const domain = lib.model.domain;
+import { default as amf, model } from "amf-client-js";
```
wjhsf, line 10:
```suggestion
import amf, { model } from "amf-client-js";
```

packages/generator/test/template-helpers-get-base-uri.test.ts
```diff
@@ -7,27 +7,25 @@
...
 
 import { getBaseUri } from "../src/templateHelpers";
 
-describe("Test getBaseUri template help function", () => {
-  before(() => wap.init());
+amf.plugins.document.WebApi.register();
 
+describe("Test getBaseUri template help function", () => {
```
wjhsf, line 16:
```suggestion
describe("getBaseUri template helper function", () => {
```
Including "test" in a test description is redundant.

packages/generator/src/renderer.ts
```diff
@@ -345,7 +394,8 @@ export async function renderTemplates(config: any): Promise<void> {
  * @returns list of operations as string
  */
 export function renderOperationList(allApis: {
-  [key: string]: WebApiBaseUnitWithEncodesModel[];
+  [key: string]: model.document.BaseUnit &
+    model.document.BaseUnitWithEncodesModel[];
```
wjhsf, line 398:
```suggestion
  [key: string]: (model.document.BaseUnit &
    model.document.BaseUnitWithEncodesModel)[];
```

packages/generator/test/renderer.test.ts
bredmond-sf: nope, wasn't working with chai-as-promised either. and yes, it does return the array, but no matter what I tried, it wasn't working properly :/

packages/generator/src/renderer.ts
```diff
@@ -126,11 +122,15 @@ export function sortApis(apis: ApiClientsInfoT[]): void {
...
-function createClient(webApiModel: WebApiBaseUnit, apiName: string): string {
+function createClient(
+  webApiModel: model.document.BaseUnit,
+  apiName: string
+): string {
   return clientInstanceTemplate(
     {
       dataTypes: getAllDataTypes(
-        webApiModel as WebApiBaseUnitWithDeclaresModel
+        webApiModel as model.document.BaseUnit &
```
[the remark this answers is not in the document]
bredmond-sf: webapi-parser package is deprecated. This union is the equivalent from the amf-client-js package. This allows us to drop that dependency.

PR #204, opened 2020-05-29
W-7643516 Move devDeps out of deps

bredmond-sf:
We have a number of packages listed in deps that should be in devDeps.


wjhsf:
Is there a story for this already, or was there just discussion of creating one?

bredmond-sf:
> Is there a story for this already, or was there just discussion of creating one?

I'm on support rotation and there was an issue opened against it (see linked issues on the side of this) so since it was quick, I just opened a PR against it

PR #209, opened 2020-06-08
Integrate diff tool with updateApis

gkhunger-sf:
Running `npm run updateApis` makes a backup of apis directory and runs a diff tool between all the apis in api-config.json. 
Diffs are written to `jsonDiff.json` file as an array of RamlDiff:
```
export type RamlDiff = {
  file: string;
  message?: string;
  diff?: NodeDiff[];
};
```
All the newly added and removed apis are listed in the results.

These changes depend on some unpublished raml-toolkit changes. Therefore, this build will fail for now.


packages/generator/test/gulpfile-helpers.test.ts
```diff
@@ -0,0 +1,201 @@
...
+
+import _ from "lodash";
+import chai from "chai";
+import chaiAsPromised from "chai-as-promised";
+import fs from "fs-extra";
+import path from "path";
+import sinon from "sinon";
+import tmp from "tmp";
+
+const expect = chai.expect;
```
wjhsf, line 21:
```suggestion
```
See above.

packages/generator/test/gulpfile-helpers.test.ts
```diff
@@ -0,0 +1,201 @@
...
+ */
+
+import * as diffProcessor from "@commerce-apps/raml-toolkit";
+import { NodeDiff } from "@commerce-apps/raml-toolkit";
+import { diffNewAndArchivedRamlFiles, RamlDiff } from "../src/gulpfileHelpers";
+import { generatorLogger } from "../src/logger";
+
+import _ from "lodash";
+import chai from "chai";
+import chaiAsPromised from "chai-as-promised";
```
wjhsf, line 15:
```suggestion
import { expect } from "chai";
```
`chai-as-promised` isn't used in this file. As a result, we don't need to call `chai.use`, so we don't need to import all of `chai`, just `expect`.

packages/generator/test/gulpfile-helpers.test.ts
```diff
@@ -0,0 +1,201 @@
...
+      family1: [
+        { assetId: "api1", fatRaml: { mainFile: "api1.raml" } },
+        { assetId: "api2", fatRaml: { mainFile: "api2.raml" } }
+      ],
+      family2: [
+        { assetId: "api3", fatRaml: { mainFile: "api3.raml" } },
+        { assetId: "api4", fatRaml: { mainFile: "api4.raml" } }
+      ]
+    };
+    nodeDiffArr.push(new NodeDiff("#/web-api/end-points/test-endpoint"));
```
wjhsf, line 46:
```suggestion
    nodeDiffArr.push(new NodeDiff("#/web-api/endpoints/test-endpoint"));
```
Choose `endpoint[s]` or `end-point[s]`, not both. 😛

packages/generator/gulpfiles/downloadRamlsTask.ts
```diff
@@ -93,13 +91,30 @@ function downloadRamlFromExchange(): Promise<void> {
...
+  const result = await diffNewAndArchivedRamlFiles(
+    config.apiBackupDir,
+    config.inputDir,
+    path.join(config.apiBackupDir, config.apiConfigFile),
+    path.join(config.inputDir, config.apiConfigFile)
+  );
+  return fs.writeFile(config.diffFile, JSON.stringify(result));
+});
+
+gulp.task("cleanup", async () => {
```
wjhsf, line 104:
```suggestion
gulp.task("cleanup", () => {
```
Either this or use async `fs`.

build-config.ts
```diff
@@ -9,6 +9,8 @@ require("dotenv").config();
 
 export default {
   inputDir: process.env.COMMERCE_SDK_INPUT_DIR || `${__dirname}/apis`,
+  apiBackupDir: process.env.COMMERCE_SDK_API_BACKUP_DIR || `${__dirname}/apiBackup`,
+  diffFile: process.env.COMMERCE_SDK_RAML_DIFF_FILE || `${__dirname}/ramlDiff.json`,
```
wjhsf, line 13: Maybe specify a `tmpDir` instead, and put both of these as static paths inside there. This would also eliminate the need for additional config options in potential future changes.

packages/generator/gulpfiles/downloadRamlsTask.ts
```diff
@@ -93,13 +91,30 @@ function downloadRamlFromExchange(): Promise<void> {
...
+  fs.removeSync(path.join(config.apiBackupDir));
+  fs.removeSync(config.diffFile);
+  return fs.moveSync(
+    path.join(config.inputDir),
+    path.join(config.apiBackupDir)
+  );
```
wjhsf, line 110:
```suggestion
  fs.removeSync(config.apiBackupDir);
  fs.removeSync(config.diffFile);
  return fs.moveSync(config.inputDir, config.apiBackupDir);
```
Wrapping a single argument in `join` is unnecessary.

packages/generator/gulpfiles/downloadRamlsTask.ts
```diff
@@ -93,13 +91,30 @@ function downloadRamlFromExchange(): Promise<void> {
...
+});
+
+gulp.task("downloadRamlFiles", async () => {
+  return downloadRamlFromExchange();
 });
```
wjhsf, line 115:
```suggestion
gulp.task("downloadRamlFiles", downloadRamlFromExchange);
```
No need to wrap the function if you're not doing anything with it.

packages/generator/src/gulpfileHelpers.ts
```diff
@@ -0,0 +1,97 @@
...
+      commonRamls.push(oldRamls[i]);
+      oldRamls[i] = null;
+    }
+  }
+
+  const result: RamlDiff[] = [];
+  // Get diffs for all the common RAMLs. If diff operation fails for a RAML,
+  // log the error, add a message to the diff object and continue with the rest
+  // of the RAML files.
+  for (const raml of commonRamls) {
```
wjhsf, line 74: If you want to execute the differencing in parallel, you could convert this to a `map`.

packages/generator/src/gulpfileHelpers.ts
```diff
@@ -0,0 +1,97 @@
+/*
```
wjhsf, line 1: Since this is a gulpfile helper file, would it make more sense to put it with the gulpfiles?

packages/generator/gulpfiles/downloadRamlsTask.ts
gkhunger-sf: I already tried that, gulp doesn't like it. It complains about missing `async`.

packages/generator/src/gulpfileHelpers.ts
gkhunger-sf: I spent some time playing around with that. It didn't work very well because I didn't want the entire process to fail in case of a failure. Also, I am transforming the result and need to link each diff with their file. I didn't wanna delay this PR because of that. I'll probably push a separate commit or create a separate PR for that.

packages/generator/src/gulpfileHelpers.ts
gkhunger-sf: I was confused about this one. I decided against it because currently all our code is inside `src` including the other helper class (templateHelpers) we have.

packages/generator/gulpfiles/downloadRamlsTask.ts
```diff
@@ -93,13 +91,30 @@ function downloadRamlFromExchange(): Promise<void> {
   });
 }
 
-gulp.task("cleanInputDir", async () => {
-  return fs.removeSync(path.join(config.inputDir));
+gulp.task("diffRamlFiles", async () => {
+  const result = await diffNewAndArchivedRamlFiles(
+    config.apiBackupDir,
+    config.inputDir,
+    path.join(config.apiBackupDir, config.apiConfigFile),
```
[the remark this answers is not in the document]
gkhunger-sf: I feel this method not knowing where the api-config.json is or what it is named makes it a little bit more flexible.

packages/generator/src/gulpfileHelpers.ts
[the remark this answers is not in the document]
gkhunger-sf: Right now it's only downloadTask gulpfile that needs a helper file. In future we might need it for the other gulpfile. I believe helper functions for both the gulpfiles can reside in the same file. What do you think?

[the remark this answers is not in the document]
unandyala: I feel having separate helper files is cleaner

packages/generator/gulpfiles/downloadRamlsTask.ts
monkey-jeff: Gulp tasks must be async.  They always return a promise.  What gaurav is doing here is fine.

packages/generator/gulpfiles/downloadRamlsTask.ts
```diff
@@ -93,13 +91,30 @@ function downloadRamlFromExchange(): Promise<void> {
...
+    path.join(config.inputDir, config.apiConfigFile)
+  );
+  return fs.writeFile(config.diffFile, JSON.stringify(result));
+});
+
+gulp.task("cleanup", async () => {
+  fs.removeSync(path.join(config.apiBackupDir));
+  fs.removeSync(config.diffFile);
+  return fs.moveSync(
```
monkey-jeff, line 107: Why is cleanup moving this directory? Naming stuff is hard but I think this task would probably be better off named "backupApis" or "preUpdateApis". cleanup can get confused with a clean task which this is not.

packages/generator/src/gulpfileHelpers.ts
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * @param oldApiDir - Existing APIs
+ * @param newApiDir - Newly downloaded APIs
+ * @param oldConfigPath - Existing config
+ * @param newConfigPath - Config for newly created APIs
+ *
+ * @returns An array of RamlDiff objects containing differences between all the
+ * old and new RAML files
+ */
+export async function diffNewAndArchivedRamlFiles(
```
monkey-jeff, line 51: this can be simplified to
```
diffNewAndArchivedRamlFiles(oldApiDir, newApiDir, configFile)
```
And perform the path + file joins in the method.

packages/generator/src/gulpfileHelpers.ts
```diff
@@ -0,0 +1,97 @@
...
+  oldApiDir: string,
+  newApiDir: string,
+  oldConfigPath: string,
+  newConfigPath: string
+): Promise<RamlDiff[]> {
+  const oldRamls = listRamlsFromConfig(oldConfigPath);
+  const newRamls = listRamlsFromConfig(newConfigPath);
+
+  const commonRamls = [];
+  // Extract all RAMLs that are part of the existing and the newly downloaded
```
monkey-jeff, line 61: Probably don't want to extract as part of this method so it is independently testable.

packages/generator/src/gulpfileHelpers.ts
monkey-jeff, line 51: This method does to much, consider splitting up.

packages/generator/src/gulpfileHelpers.ts
```diff
@@ -0,0 +1,97 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import * as path from "path";
+import _ from "lodash";
+
+import { diffRaml, NodeDiff } from "@commerce-apps/raml-toolkit";
+import { generatorLogger } from "./logger";
+
+export type RamlDiff = {
```
monkey-jeff, line 13: Should this type be defined bt the raml-toolkit instead?

packages/generator/src/gulpfileHelpers.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+  const removedRamls: RamlDiff[] = oldRamls
+    .filter(r => r)
+    .map(r => ({ file: r, message: "This RAML has been removed" }));
+  const addedRamls: RamlDiff[] = newRamls
+    .filter(r => !commonRamls.includes(r))
+    .map(r => ({ file: r, message: "This RAML has been added recently" }));
+
+  return result.concat(removedRamls, addedRamls);
```
monkey-jeff, line 96: How do we know if there are breaking changes?

packages/generator/src/gulpfileHelpers.ts
gkhunger-sf: I feel it is very specific to this use case. Why do you feel raml-toolkit is a better place for this?

monkey-jeff: Because the eventual goal is to remove all of this logic from here :-p

build-config.ts
gkhunger-personal: We are gonna be using this every release, I feel a predictable location for this file would be convenient. I don't see any major advantage of putting the file in a tmp dir. I am going to skip this, if you don't mind.

gkhunger-sf: We are gonna be using this every release, I feel a predictable location for this file would be convenient. I don't see any major advantage of putting the file in a tmp dir. I am going to skip this, if you don't mind.

gkhunger-sf: This makes sense. Will make the change.

packages/generator/gulpfiles/downloadRamlsTask.ts
```diff
@@ -65,11 +65,29 @@ async function search(): Promise<RestApi[]> {
   });
 }
 
+gulp.task("diffRamlFiles", async () => {
+  const result = await diffNewAndArchivedRamlFiles(
+    config.apiBackupDir,
+    config.inputDir,
+    path.join(config.apiConfigFile)
```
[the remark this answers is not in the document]
gkhunger-sf: aaah. missed this one.

packages/generator/src/downloadRamlsGulpTaskHelpers.ts
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * @param array1 - One of the arrays being compared
+ * @param array2 - The other array being compared
+ * @param common - Return all the elements that are common if true, all the
+ * elements that are not common otherwise
+ *
+ * @returns An array of RAML files that are common or not common in both the
+ * arrays
+ */
+function compareArrays(
```
wjhsf, line 44: Calling this multiple times with the same args seems inefficient. Maybe do something like this?

```typescript
function split<T>(leftArr: T[], rightArr: T[]): {
  left: T[];
  right: T[];
  common: T[];
} {
  const left = new Set(leftArr);
  const right = new Set(rightArr);
  const common = new Set<T>();
  left.forEach(val => {
    if (right.has(val)) {
      left.delete(val);
      right.delete(val);
      common.add(val);
    }
  });
  return {
    left: [...left],
    right: [...right],
    common: [...common]
  };
}
```

packages/generator/test/download-ramls-gulp-task-helpers.test.ts
```diff
@@ -0,0 +1,178 @@
+/*
```
wjhsf, line 1: Test file names should be camel case.

gkhunger-sf: I don't know what the rationale behind naming them kebab-case is but I would like to know that before changing the names. And that should be a separate PR.

bredmond-sf: We started with the more widely accepted practice of kebab-case, but changed to align with the LWC. It's what we call out in our best practices now. We didn't make it high priority to rename tests, but decided to just update them as we go.

gkhunger-sf: Thanks for clarifying @bredmond-sf, will change them.

packages/generator/src/downloadRamlsGulpTaskHelpers.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  return ramls;
+}
+
+/**
+ * Compares the arrays and returns all the common and exclusive elements.
+ *
+ * @param leftArr - One of the arrays being compared
+ * @param rightArr - The other array being compared
+ * elements that are not common otherwise
```
wjhsf, line 38:
```suggestion
```

packages/generator/src/downloadRamlsGulpTaskHelpers.ts
```diff
@@ -0,0 +1,131 @@
...
+    for (const api of config[apiFamily]) {
+      ramls.push(path.join(api.assetId, api.fatRaml.mainFile));
+    }
+  }
+
+  return ramls;
+}
+
+/**
+ * Compares the arrays and returns all the common and exclusive elements.
```
wjhsf, line 34:
```suggestion
 * Compares two arrays and returns their intersection and symmetric difference.
```
Fancy math terms!

gkhunger-sf: Sounds too technical but happy to change if it makes you happy 😁

PR #212, opened 2020-06-16
Ka/cache bug.w 7649264

kalagarraj:
Make fetch happen depends on mini pass streams package. We found an issue where the large responses are not cached and the requests are failing silently. The failure happens to be around concatenating the response stream. 

The cache manager used for Redis do not have this issue. Using the same cache manager with plain old Javascript Map API based Keyv store as the default cache manager did not result in the failure as well. 

changes
Default cacache is replaced with Keyv Cache Manager with quick-lru cache.

3PP filed for quick-lru https://gus.lightning.force.com/lightning/r/ADM_Third_Party_Software__c/a0qB0000000hBDRIA2/view


packages/core/src/base/client.ts
```diff
@@ -42,9 +43,7 @@ export class ClientConfig {
 
 const DEFAULT_CLIENT_CONFIG: ClientConfig = {
   // Enables cacache for local caching in temp dir by default, unsafeCleanup == rm -rf on exit
-  cacheManager: new DefaultCache(
-    tmp.dirSync({ prefix: "cache-", unsafeCleanup: true }).name
-  ),
+  cacheManager: new CacheManagerKeyv( { keyvStore: new Map() }),
```
jbegleiter-sf, line 46: Does keyv still do expiry the correct way using this?

kalagarraj: Setting expiry is still a function of the cache-manager. This uses the cache manager @bredmond-sf wrote for Redis.

packages/core/src/base/client.ts
```diff
@@ -42,9 +43,7 @@ export class ClientConfig {
 
 const DEFAULT_CLIENT_CONFIG: ClientConfig = {
   // Enables cacache for local caching in temp dir by default, unsafeCleanup == rm -rf on exit
```
unandyala, line 45: change the comment

packages/core/src/base/client.ts
bredmond-sf: Discussed and let's use quick-lru for the store.

kalagarraj: Done

packages/core/src/base/client.ts
kalagarraj: Done

packages/core/package.json
```diff
@@ -121,6 +121,7 @@
   "dependencies": {
     "@commerce-apps/raml-toolkit": "^0.4.2",
     "@keyv/redis": "^2.1.1",
+    "quick-lru" : "^5.1.1",
```
kalagarraj, line 124: 3PP for quick-lru https://gus.lightning.force.com/lightning/r/ADM_Third_Party_Software__c/a0qB0000000hBDRIA2/view

packages/generator/README.md
```diff
@@ -112,7 +112,27 @@ Autocomplete will also show the available properties of the data returned by SDK
...
+The SDK currently supports two types of caches - In-memory and Redis. Both the implementations respect [standard cache headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control). If you would like to use another type of cache, you can write your own implementation of the [CacheManager](../core/src/base/cacheManager.ts). You can refer to this [default cache manager](../core/src/base/cacheManagerKeyv.ts) when designing your implementation.  
+
+### Cache storage adapter
+
+The default cache storage limits to 10,000 distinct entities before applying a simple [least recently used](https://en.m.wikipedia.org/wiki/Cache_replacement_policies#Least_Recently_Used_.28LRU.29) policy for cache replacement. You can change the limit by creating a [quick-lru](https://www.npmjs.com/package/quick-lru) storage adapter.
+```javascript
+import { CacheManagerKeyv } from "@commerce-apps/core";
+import { QuickLRU } from "quick-lru";
+
+const cacheManagerKeyv = new CacheManagerKeyv({ keyvStore: new QuickLRU({ maxSize: 50000 })});
```
monkey-jeff, line 124: We might want to call out more explicitly that this is optional to pass and not required.

kalagarraj: maxSize is required https://github.com/sindresorhus/quick-lru#readme

kalagarraj: Looks like their readme is wrong. Their code requires options.maxSize > 0

monkey-jeff: I was talking about from our perspective it can me make clearer what happens when you don't specify a cache manager at all.   Again not blocking definitely can close.

PR #217, opened 2020-06-25
Add a function to export headers object as map

gkhunger-sf:


packages/core/src/base/staticClient.ts
```diff
@@ -93,6 +93,30 @@ export const logResponse = (response: Response): void => {
...
+ * headers expects them to be comma separated strings.
+ *
+ * @param headers - a minipass-fetch Headers object
+ *
+ * @returns a map of headers
+ */
+export function exportHeadersAsMap(headers: Headers) {
+  const res = {};
+  const keysIterator: HeadersIterator = headers.keys();
+  let key = keysIterator.next();
```
kalagarraj, line 111: Lodash have many functions to array operations that can be used

packages/core/src/base/staticClient.ts
```diff
@@ -93,6 +93,30 @@ export const logResponse = (response: Response): void => {
...
+ * minipass-fetch Headers class has a function to convert a Headers object into
+ * a map. But that function creates an array of values for each header. 
+ * `http-cache-semantics`, a package used by `make-fetch-happen` to manipulate
+ * headers expects them to be comma separated strings.
+ *
+ * @param headers - a minipass-fetch Headers object
+ *
+ * @returns a map of headers
+ */
+export function exportHeadersAsMap(headers: Headers) {
```
kalagarraj, line 108: Do we need this function exported?

packages/core/src/base/staticClient.ts
```diff
@@ -93,6 +93,30 @@ export const logResponse = (response: Response): void => {
...
+  const res = {};
+  const keysIterator: HeadersIterator = headers.keys();
```
kalagarraj, line 110: check for null headers, u can use headers?.keys()

packages/core/src/base/staticClient.ts
gkhunger-sf: For tests, not otherwise

packages/core/src/base/staticClient.ts
```diff
@@ -93,6 +93,30 @@ export const logResponse = (response: Response): void => {
...
+  let key = keysIterator.next();
+  while (!key.done) {
```
monkey-jeff, line 112: This works, but why not something like

```
export function exportHeadersAsMap(headers: Headers) {
  const res = {};
  const headersIterator: HeadersIterator = headers.entries();
  for (const [key, value] of headersIterator) {
    res[key] = value;
  }
  return res;
}
```
?

Just think it is cleaner to read

packages/core/src/base/staticClient.ts
wjhsf, line 108: The function name name is inaccurate:
1) This function works with any iterable, not just headers.
2) This function returns an Object, not a Map.

packages/core/src/base/staticClient.ts
wjhsf: The cleanest solution would actually be to just use `Object.fromEntries(headers)`, but that's not available in node 10. Lodash does provide a convenient alternative, though: `_.fromPairs([...headers])`.

Generally, I agree that we should be trying to avoid lodash whenever possible, but I think in this case it does provide a clean solution that allows us to avoid writing our own custom function.

gkhunger-sf:
> Look good, couple of comments but nothing blocking.
> 
> One thing I would like to see though is a negative test that fails for the existing implementation and succeeds in the new one. Without that I don't think we can call the bug closed.

@monkey-jeff I believe integration tests are a better place for that test. I'm already looking into them to figure out why they aren't currently failing with this issue.

packages/core/test/staticClient/headers.test.ts
```diff
@@ -258,4 +259,30 @@ describe("Base Client headers", () => {
       expect(nock.isDone()).to.be.true;
     });
   });
+
+  describe("getObjectFromResponse", () => {
+
+    it("should return a map of headers", () => {
```
wjhsf, line 265: `Map` in JavaScript is a specific class. These tests only check plain objects.

wjhsf, line 265:
```suggestion
    it("returns a map of headers", () => {
```
As per best practices style guide.

packages/core/test/staticClient/headers.test.ts
```diff
@@ -258,4 +259,30 @@ describe("Base Client headers", () => {
...
+      const headersMap = {
+        authorization: "Bearer token",
+        connection: "close"
+      }
+      const headers = new Headers(headersMap);
+
+      expect(exportHeadersAsMap(headers)).to.deep.equal(headersMap);
+    });
+
+    it("should return a comma separated list if a header has multiple values", () => {
```
wjhsf, line 275:
```suggestion
    it("returns a comma separated list if a header has multiple values", () => {
```
As per best practices style guide.

packages/core/test/staticClient/headers.test.ts
```diff
@@ -258,4 +259,30 @@ describe("Base Client headers", () => {
...
+      const headersMap = {
+        accept: "text/plain, text/html"
+      }
+      const headers: Headers = new Headers(headersMap);
```
wjhsf, line 279:
```suggestion
      const headers = new Headers(headersMap);
```
Type inferencing makes the annotation unnecessary.

packages/core/test/staticClient/headers.test.ts
```diff
@@ -258,4 +259,30 @@ describe("Base Client headers", () => {
...
+
+      expect(exportHeadersAsMap(headers)).to.deep.equal(headersMap);
+    });
+
+    it("should return an empty map if no headers are passed", () => {
```
wjhsf, line 284:
```suggestion
    it("returns an empty map if no headers are passed", () => {
```
As per best practices style guide.

packages/core/test/staticClient/headers.test.ts
```diff
@@ -258,4 +259,30 @@ describe("Base Client headers", () => {
       expect(nock.isDone()).to.be.true;
     });
   });
+
+  describe("getObjectFromResponse", () => {
```
wjhsf, line 263:
```suggestion
  describe("exportHeadersAsMap", () => {
```
Missed a spot! 😉

packages/core/src/base/staticClient.ts
bredmond-sf:
```
const res = {};
headers.forEach((val, key, obj) => res[key] = obj.get(key));
return res;
```

packages/core/src/base/staticClient.ts
bredmond-sf: Also, this should probably be headersToObjectLiteral() since it doesn't return a Map.

packages/core/src/base/staticClient.ts
bredmond-sf: > The cleanest solution would actually be to just use `Object.fromEntries(headers)`, but that's not available in node 10. Lodash does provide a convenient alternative, though: `_.fromPairs([...headers])`.
Those give you the same result as just calling headers.raw(). The problem this addresses is calling 'get' which does a join on the value before returning.

gkhunger-sf: `_.fromPairs()` works. `.raw()` only creates arrays of values whereas `...headers` creates arrays of `key + value` which is what `fromPairs` needs to create an object of key value strings.

PR #218, opened 2020-06-25
Generation of sdk documentation from rendered templates

dbarbato-sf:
Enables generation of documentation for the sdk.
The documentation will be hosted by github pages. Github pages requires the documents to live on the master branch under /docs
As a part of reworking this repo, there will be some changes required to document generation, these changes will be captured as a part of @bredmond-sf 's spike doc.
Work remaining outside of this PR to complete the story includes:

- [ ] 1. Setting up and configuring github pages.
- [ ] 2. Updating @bredmond-sf 's doc.
- [ ] 3. Creating a quick start guide for generating doc on other projects.
- [ ] 4. Update release docs.


package.json
```diff
@@ -20,6 +20,7 @@
     "test:all": "npm run test && npm run test:integration",
     "test:all:redis": "npm run test && npm run test:integration:redis",
     "test:all:redis:debug": "npm run test && npm run test:integration:redis:debug",
+    "doc": "lerna run doc",
```
bredmond-sf, line 23: Can we skip adding this? Just one less thing to consider when we remove lerna.

packages/generator/package.json
```diff
@@ -23,6 +23,7 @@
     "build": "npm run renderTemplates && tsc && eslint dist --ext .ts,.js --quiet --fix",
     "test": "nyc mocha \"test/**/*.test.ts\" --reporter=xunit --reporter-options output=../../reports/generator.xml",
     "test:debug": "nyc mocha \"test/**/*.test.ts\"",
+    "doc": "typedoc --mode modules --out ../../docs renderedTemplates/** --tsconfig tsconfig.json --external-modulemap  \".*/renderedTemplates/([\\w\\-_]+)/\"",
```
bredmond-sf, line 26: Couple things here:
- Why docs at the root instead of just in this package?
- Do we need to check in the rendered docs?
  - If so, how do we keep them in sync with the code and make sure they're updated? Triggered by the build command?
  - If not, this should npm run renderTemplates

monkey-jeff:
What is your vision for a process to updating this?

package.json
dbarbato-sf: I guess we could, I was imagining the docs would be updated by the designated "release engineer", I was going to update the release doc to state npm run doc should kick off after npm run build, I put this in here for convenience.

I can either remove it or link this directly to the command in the generator package.json (or just leave it here for now).
I don't think it's a big deal either way, but happy to go which ever direction you like.

dbarbato-sf:
> What is your vision for the process of updating the docs?
@monkey-jeff for now, In the release notes I'm going to add a line to update and commit the docs the designated release engineer runs the build.

PR #226, opened 2020-07-03
Updated the mercury APIs and standards

dbarbato-sf:


monkey-jeff:
Why are we updating this out of cycle?

dbarbato-sf:
I was unaware this was on a cycle, this will get done prior to release anyway, why would this be an issue?

I’d updated my other branch and pulled these in to that, so figured rather than reverting I’d get this branch in ahead as it’d need to go in sometime anyway.

I can revert and close this if that’s preferable, but would like to understand why this adds complexity.

monkey-jeff:
Its just often we check to see if there are APIs updated by running update APIs in the go no go.  Definitely not an issue to do it out of cycle.  But we need to make sure that we are aware of those changes and include them in the list of changes.  This really isn't a problem I don't think.  Just no one has done it before.

PR #229, opened 2020-07-09
Final Review: Removing T postfix from types and adding a "namespace"

dbarbato-sf:


packages/generator/src/commerceTemplateHelper.ts
```diff
@@ -0,0 +1,96 @@
...
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const templateHelpers = require("../src/templateHelpers");
+
+/**
+ * Given a type, prefix it with the namespace "types."
+ * @param type - in string format to be prefixed
+ * @returns the type prefixed by types.
+ */
+export function addNamespacePrefixToType(type: string): string {
+  if (!type) {
```
monkey-jeff, line 19: If type is untruthy we return type? Is this the behavior we want here?

packages/generator/src/commerceTemplateHelper.ts
```diff
@@ -0,0 +1,96 @@
...
+    return type;
+  }
+  const prefix = "types.";
```
monkey-jeff, line 22: Somewhat a mid method magic value.  Might want this to be a true constant at the top and call it TYPE_PREFIX or something clear in this use case.

packages/generator/src/commerceTemplateHelper.ts
```diff
@@ -0,0 +1,96 @@
...
+  const types = type.split("|");
```
monkey-jeff, line 23: I am confused about the flow here? Is this happening after the rendering of the list of types?  Wouldn't this be simpler if it happened prior to that union type listing?

packages/generator/src/commerceTemplateHelper.ts
```diff
@@ -0,0 +1,96 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const templateHelpers = require("../src/templateHelpers");
+
+/**
+ * Given a type, prefix it with the namespace "types."
+ * @param type - in string format to be prefixed
+ * @returns the type prefixed by types.
+ */
+export function addNamespacePrefixToType(type: string): string {
```
monkey-jeff, line 18: This method seems to have a lot of logic copied from other methods.  Is there a way to reuse that logic and simplify this method instead of repeating some of the logic here?

packages/generator/src/commerceTemplateHelper.ts
```diff
@@ -0,0 +1,96 @@
...
+  }
+  const prefix = "types.";
+  const types = type.split("|");
+
+  let namespaceTypes = types.filter(checkType =>{
+    return checkType.trim() != '';
+      
+  });
+  namespaceTypes = namespaceTypes.map(checkType => {
+    if (["void", "Object", "object"].includes(checkType)) {
```
monkey-jeff, line 30:
```suggestion
    if (["void", "object"].includes(checkType.toLowerCase())) {
```

packages/generator/src/commerceTemplateHelper.ts
```diff
@@ -0,0 +1,96 @@
...
+  const namespaceType = namespaceTypes.join(" | ");
+  return namespaceType;
+}
+
+/**
+ * Given an array in string format, prefix each contained type with the namespace "types." 
+ * @param type - in the format Array\<Type1|Type2\>
+ * @returns the array with each element prefixed with the namespace, eg Array\<types.Type1 | types.Type2\>
+ */
+export function addNamespacePrefixToArray(array: string): string {
```
monkey-jeff, line 44: Is this again parsing already rendered code instead of parsing that raw type?

packages/generator/src/commerceTemplateHelper.ts
```diff
@@ -0,0 +1,96 @@
...
+  if (!array) {
```
monkey-jeff, line 45: Same comment as before. If untruthy do we really want to return it as is?

packages/generator/src/commerceTemplateHelper.ts
```diff
@@ -0,0 +1,96 @@
...
+): string {
+  if (
+    request != null &&
+    request.payloads != null &&
+    request.payloads.length > 0
+  ) {
+    const payloadSchema: model.domain.Shape = request.payloads[0].schema;
+    if (payloadSchema instanceof model.domain.ArrayShape) {
+      return addNamespacePrefixToArray(
+        templateHelpers.getRequestPayloadType(request)
```
monkey-jeff, line 73: I think this is the core of the issue I am referring to.  By using the existing template helper I think that we are getting the rendered output...having to deconstruct it and then reconstruct it.  We should probably have an intermediary method that both this and the getRequestPayloadType can utilize for there respective renderings.

packages/generator/templates/operations.ts.hbs
```diff
@@ -34,9 +34,9 @@
         },
         retrySettings?: OperationOptions,
         headers?: { [key: string]: string }{{#or (is method "patch") (is method "post") (is method "put")}},
-        body: {{{getRequestPayloadType request}}}{{/or}}
+        body: {{{getRequestPayloadTypeWithNamespace request}}}{{/or}}
```
monkey-jeff, line 37: Another option is to have this be 2 helpers.  One that calls another within this template.  Something like `{{{addNamespace (getRequestPayloadType request)}}}` This would also help to ensure DRYer code for this process.

packages/generator/templates/operations.ts.hbs
```diff
@@ -109,10 +109,10 @@
         },
         retrySettings?: OperationOptions,
         headers?: { [key: string]: string }{{#or (is method "patch") (is method "post") (is method "put")}}, 
-        body: {{{getRequestPayloadType request}}}{{/or}}
+        body: {{{getRequestPayloadTypeWithNamespace request}}}{{/or}}
       },
       rawResponse?: boolean,
-    ): Promise<Response | {{getReturnPayloadType .}}> {
+    ): Promise<Response | {{getReturnPayloadTypeWithNamespace .}}> {
```
monkey-jeff, line 115: Do you know the difference between {{{ and {{ 
?

packages/generator/src/commerceTemplateHelper.ts
```diff
@@ -0,0 +1,96 @@
+/*
```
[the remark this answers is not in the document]
dbarbato-sf: It's the wrapper commerce is using around template helpers. The thinking was that if we needed to modify any other default functionality, it'd also happen in this file.

packages/core/test/auth-helper.test.ts
```diff
@@ -9,7 +9,7 @@ import { ShopperToken, stripBearer } from "../src/base/authHelper";
 import { expect, default as chai } from "chai";
 import chaiAsPromised from "chai-as-promised";
 
-import { CustomerT } from "../../generator/dist/customer/shopperCustomers/shopperCustomers.types";
+import { Customer } from "../../generator/dist/customer/shopperCustomers/shopperCustomers.types";
```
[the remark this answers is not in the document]
dbarbato-sf: That's not merged yet so it needs to be here to allow this to merge if it goes before :)

packages/generator/src/commerceTemplateHelper.ts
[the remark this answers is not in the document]
dbarbato-sf: Will rename customTemplateHelpers

packages/generator/src/commerceTemplateHelper.ts
dbarbato-sf: Ensure trim happens at top.

packages/generator/src/renderer.ts
```diff
@@ -37,6 +37,7 @@ import {
   getPascalCaseName,
   formatForTsDoc
 } from "./templateHelpers";
+import { addNamespace } from "./commonTemplateHelper";
```
wjhsf, line 40: Why is this in its own file and not with the other template helpers?

packages/generator/test/commonTemplateHelper.test.ts
```diff
@@ -0,0 +1,69 @@
...
+      .and.equal("Array<types.Foo | types.Baa>");
+  });
+
+  it("should correctly parse three elements in an array", () => {
+    expect(addNamespace("Array<Foo | Baa | Bar>", "types"))
+      .to.be.a("string")
+      .and.equal("Array<types.Foo | types.Baa | types.Bar>");
+  });
+
+  it("should correctly parse no elements in an array", () => {
```
wjhsf, line 62:
```suggestion
  it("should throw an error if no elements in an array", () => {
```
"correctly parse" sounds like it should succeed, not throw.

packages/generator/test/commonTemplateHelper.test.ts
```diff
@@ -0,0 +1,69 @@
...
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+import { expect } from "chai";
+import { addNamespace } from "../src/commonTemplateHelper";
+
+describe("When adding namespaces to individual types", () => {
+  it("Should prefix the types namespace when provided with any string ", () => {
```
wjhsf, line 12:
```suggestion
  it("prefixes any string with a namespace ", () => {
```
Per our best practices / style guide, test descriptions should be `it("does a thing")`, not `it("should do a thing")`.

packages/generator/src/commonTemplateHelper.ts
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Given an individual type or an array of types in the format Array\<Foo | Baa\>
+ * will return either the type prefixed by the namespace, or the Array with each type prefixed
+ * eg. Array\<types.Foo | types.Baa\>
+ *
+ * @param content - to be parsed for types to prefix with a namespace
+ * @param namespace - to be prefixed to types
+ * @returns the content prefixed with the namespace
+ */
+export function addNamespace(content: string, namespace: string): string {
```
wjhsf, line 17: Looking at the templates, is the `namespace` provided ever not `"types"`? Will it ever be different? The namespace could be hard-coded or made an optional parameter (although the latter is a small bit of extra work due to Handlebars).

packages/generator/src/commonTemplateHelper.ts
```diff
@@ -0,0 +1,64 @@
...
+    throw new Error("Invalid namespace");
+  }
+
+  // if the content is an array, extract all of the elements
+  let types = content;
+  let arrayType = false;
+  if (content.match("^Array") !== null) {
+    types = content.replace(/Array|<|>/g, "");
+    arrayType = true;
+  }
```
wjhsf, line 33:
```suggestion
  const matched = content.match(/^Array<(.*?)>$/);
  const arrayType = !!matched;
  const types = matched?.[1] || content;
```
1) `content.match()` should take a RegExp, not a string.
2) Since you're using RegExp, take advantage of match groups!

packages/generator/src/commonTemplateHelper.ts
```diff
@@ -0,0 +1,64 @@
...
+  // Get a handle on individual types
+  const typesToProcess = types.split("|");
+  const namespaceTypes: string[] = [];
+
+  // for each type
+  typesToProcess.forEach(checkType => {
+    // trim the fat
+    const actualType = checkType.trim();
```
wjhsf, line 41:
```suggestion
  const typesToProcess = types.split(/\s*|\s*/);
  const namespaceTypes: string[] = [];

  // for each type
  typesToProcess.forEach(checkType => {
```
`types.split()` can take a RegExp that makes calling `trim` unnecessary.

packages/generator/src/commonTemplateHelper.ts
```diff
@@ -0,0 +1,64 @@
...
+    // check if there's an actual type present
+    if (actualType === "") {
+      throw new Error("Empty type found");
+    }
+
+    // void and object types don't get a namespace
+    if (["void", "object"].includes(actualType.toLocaleLowerCase())) {
```
wjhsf, line 48: Are these guaranteed to be the only two types we don't want to namespace? We'll never encounter `null` or anything else?

wjhsf, line 48:
```suggestion
    if (["void", "object"].includes(actualType)) {
```
Converting to lowercase is unnecessary. Types are case-sensitive. (There technically is an `Object` type, but we shouldn't ever encounter it.)

packages/generator/src/commonTemplateHelper.ts
monkey-jeff: Originally had it that way.  Actually talking it over with Daniel we changed it to be this way.  Basically, I think this is a typescript/javascript utility function and others might want to prefix it differently, like DTO or something.  But more than that I wanted the context of what is being added in the handlebars template and not in the javascript code.

packages/generator/src/commonTemplateHelper.ts
monkey-jeff: I asked this same question to Daniel.  As far as we know these are the only two we've seen and both agreed that we should only have it there for now if we know about it.  If another pops up we'll write a test and add that as well.

monkey-jeff: We did encounter object and Object in the list...which is why this was added.

packages/generator/package.json
```diff
@@ -165,6 +165,7 @@
     "conditional-type-checks": "^1.0.5",
     "depcheck": "^0.9.2",
     "dotenv": "^8.2.0",
+    "eslint": "^7.4.0",
```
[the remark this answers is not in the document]
dbarbato-sf: It shouldn't make a difference, and @monkey-jeff has an incoming PR that will resolve this anyway.

packages/generator/src/commonTemplateHelper.ts
```diff
@@ -0,0 +1,64 @@
...
+      namespaceTypes.push(`${namespace}.${actualType}`);
+    }
+  });
+
+  // reconstruct the passed in type with the namespace
+  const processedTypes = namespaceTypes.join(" | ");
+
+  // Re-add Array if required
+  if (arrayType) {
+    return "Array<".concat(processedTypes).concat(">");
```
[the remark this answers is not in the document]
dbarbato-sf: Why yes it is, thank you for noticing!

packages/generator/src/renderer.ts
dbarbato-sf: I think @monkey-jeff answered this elsewhere, but, the other helpers are being moved out to the raml-toolkit, this is something that is not likely going to be reused by other SDKs. It also provides a way for other consumers to understand how the framework can be extended.

packages/generator/src/commonTemplateHelper.ts
wjhsf: In that case, our usage of `Object` should be removed. Using `Object` the #1 thing on TypeScript's official [list of do's and don'ts](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html).

We use `Object` in a single location: https://github.com/SalesforceCommerceCloud/commerce-sdk/blob/0e9a36810365b31d2f115bb37d289e1b1d78107e/packages/generator/src/templateHelpers.ts#L94

Changing it to return `object` instead will bring us in line with best practices and will make the SDK more consistent.

The one caveat is that `object` will complain about property access, but `Object` will not. But based on our other usage of `object`, this doesn't seem to be a major concern.

PR #234, opened 2020-07-12
Fixed the losing of images after the refactor

monkey-jeff:
Seems as if during the refactor, images were lost. Added them back in and fixed some of the document generation.


README.md
```diff
@@ -106,18 +106,18 @@ helpers.getShopperToken(config, { type: "guest" }).then(async (token) => {
 When using an IDE such as VSCode, the autocomplete feature lets you view the available method and class definitions, including parameters.
 ​
 
-![Autocomplete](https://github.com/SalesforceCommerceCloud/commerce-sdk/raw/master/images/Autocomplete.jpg?raw=true 'Autocomplete')
+![Autocomplete](./images/Autocomplete.jpg?raw=true 'Autocomplete')
```
bredmond-sf, line 109: This will be broken on the npm page, that's why it was an absolute URL

wjhsf: I think the links were broken because of the monorepo structure. Now that we've gotten rid of it, I think npm can handle relative links.

(It might not. I remember that there were npm vs monorepo issues, but I'm not 100% sure if this was one of them.)

monkey-jeff: It might be.  Someone can followup on this but I don't think this is the time to do that investigation.  Right now it is broken completely so let us fix that first.

README.md
```diff
@@ -106,18 +106,18 @@ helpers.getShopperToken(config, { type: "guest" }).then(async (token) => {
 When using an IDE such as VSCode, the autocomplete feature lets you view the available method and class definitions, including parameters.
```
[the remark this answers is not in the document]
monkey-jeff: I think we should probably do a full doc review at some point.

PR #235, opened 2020-07-13
Remove handlebars helpers as they have been moved to raml-toolkit

gkhunger-sf:
* Remove all the handlebars helpers as they have now been moved to raml-toolkit
* Move all the handlebars config to a new file - handlebarsConfig.ts

PR fails currently because of dependence on raml-toolkit


templates/dto.ts.hbs
```diff
@@ -1,5 +1,5 @@
 {{#each .}}
     {{#if (isTypeDefinition .)}}
-        export type {{getValue name}} = {{> dtoPartial typeDto=. }}
```
wjhsf, line 3: Was this intentional?

PR #241, opened 2020-07-16
Update docs

wjhsf:
Noticed that the docs still have types with T postfix.


saivenkatesh48:
Can i integrate this sdk for mobile apps, which platform it supports to work for Mobile platform?
can you please suggest me the process of integrating the sdk fro mobile.

PR #238, opened 2020-07-15
Update updateApis to use raml-toolkit

wjhsf:
Changed updateApis command to use raml-toolkit and removed unnecessary code.
`npm run updateApis` should have identical functionality to before this PR, except I changed the names of some folders.

Note: This passes tests, but `npm run updateApis` will fail until the next release of raml-toolkit.

PR #243, opened 2020-07-17
Use new names of raml-toolkit handlebarsAmfHelpers

gkhunger-sf:


wjhsf:
Note: This PR technically breaks `npm run updateApis`, but that's okay, because it will be immediately followed by #238, which fixes it.

PR #249, opened 2020-07-22
Fix data type resolution for nested types

monkey-jeff:
Due to method calls modifying the original objects..order mattered.  Needed to refactor to keep this in mind.  (We should update the methods to not modify original objects)


src/renderer.ts
```diff
@@ -137,10 +137,15 @@ function createClient(
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   apiMetadata: { [key: string]: any }
 ): string {
+  // TODO: getAllDataTypes should NOT modify original object
```
dbarbato-sf, line 140: Can we add a story for this?

PR #255, opened 2020-07-27
Update typescript to 3.9

wjhsf:


package.json
```diff
@@ -189,6 +189,7 @@
     "typedoc": "^0.17.8",
     "typedoc-plugin-external-module-map": "^1.2.1",
     "typedoc-plugin-nojekyll": "^1.0.1",
+    "typescript": "^3.9.7",
```
kalagarraj, line 192: you do not need this PR. You have a core PR that takes care of this..

wjhsf: TypeScript is a dependency of this project. We should not rely on other projects to provide it. (That's one of the reasons why we use depcheck.)

PR #259, opened 2020-08-24
Feature/new pipeline.code only

monkey-jeff:
Just the code changes for #257 

Review here but merge there.  Due to api refactor the change list was far too much.


src/lib/templateHelpers.test.ts
```diff
@@ -0,0 +1,197 @@
...
+  });
+
+  it("Will not prefix object with a namespace", () => {
+    expect(addNamespace("object", "types"))
+      .to.be.a("string")
+      .and.equal("object");
+  });
+
+  it("Will not prefix Object with a namespace", () => {
+    expect(addNamespace("object", "types"))
```
kalagarraj, line 35: are you trying to test "Object"?

src/lib/templateHelpers.test.ts
```diff
@@ -0,0 +1,197 @@
...
+    expect(addNamespace("foo", "types"))
+      .to.be.a("string")
+      .and.equal("types.foo");
+  });
+
+  it("Will not prefix void with a namespace", () => {
+    expect(addNamespace("void", "types")).to.be.a("string").and.equal("void");
+  });
+
+  it("Will not prefix object with a namespace", () => {
```
kalagarraj, line 28: I see duplicate tests with different description, Are you trying to test for case insensitive string "object"?

src/lib/templateHelpers.test.ts
```diff
@@ -0,0 +1,197 @@
...
+      .and.equal("object");
+  });
+
+  it("Will not prefix Object with a namespace", () => {
+    expect(addNamespace("object", "types"))
+      .to.be.a("string")
+      .and.equal("object");
+  });
+
+  it("Throws an error when called with undfined content", () => {
```
kalagarraj, line 40: undfined typo

src/lib/templateHelpers.test.ts
```diff
@@ -0,0 +1,197 @@
...
+    expect(() => addNamespace(null, "types")).to.throw("Invalid content");
+  });
+
+  it("Throws an error when called with null content", () => {
+    expect(() => addNamespace(undefined, "types")).to.throw("Invalid content");
+  });
+
+  it("Throws an error when the type is not a valid string", () => {
```
kalagarraj, line 48: add few more tests for empty string, string with only spaces

src/lib/templateHelpers.test.ts
```diff
@@ -0,0 +1,197 @@
...
+});
+
+describe("When adding namespaces to elements in a complex content (an array)", () => {
+  it("Prefixes each element with a namespace", () => {
+    expect(addNamespace("Array<Foo | Baa>", "types"))
+      .to.be.a("string")
+      .and.equal("Array<types.Foo | types.Baa>");
+  });
+
+  it("Correctly parses three elements in an array", () => {
```
kalagarraj, line 60: Add a test for zero and one element as well with a namespace

src/lib/templateHelpers.ts
```diff
@@ -4,35 +4,66 @@
...
- * @returns true if the parameter is a common parameter
- */
-export const isCommonQueryParameter = (property: string): boolean =>
-  property
-    ? commonParameterPositions.queryParameters.includes(property.toString())
-    : false;
+  // if the content is an array, extract all of the elements
+  const matched = content.match(/^Array<(.*?)>$/);
+  const arrayType = !!matched;
+  const types = matched?.[1] || content;
```
kalagarraj, line 34: potential index out of bounds error.. do u want to check if the length is greater than 1 before getting the index 1

src/lib/templateHelpers.ts
```diff
@@ -103,5 +98,26 @@ export const formatForTsDoc = (str: string): string => {
...
+/**
+ * Checks if a path parameter is one of the set that are configurable at the client level
+ *
+ * @param property - The string name of the parameter to check
+ *
+ * @returns true if the parameter is a common parameter
+ */
+export const isCommonPathParameter = (property: string): boolean =>
+  property
+    ? commonParameterPositions.pathParameters.includes(property.toString())
```
kalagarraj, line 110: Why do we need to call toString() on string?

src/lib/templateHelpers.test.ts
monkey-jeff: This test was moved wholesale.  Originally written by @dbarbato-sf

src/lib/templateHelpers.test.ts
monkey-jeff: This was moved wholesale. I would prefer not to much else beyond the refactor to keep it focused.

src/lib/templateHelpers.test.ts
monkey-jeff: Same comment

src/lib/templateHelpers.ts
monkey-jeff: This was also copied over.  This might be an issue but we should confirm.  Also written by @dbarbato-sf

src/lib/templateHelpers.ts
monkey-jeff: Don't know, didn't change this.

src/lib/templateHelpers.ts
wjhsf: JavaScript doesn't throw if you access an unset index, it just returns `undefined`. This is safe.

PR #261, opened 2020-09-02
Added typedoc plugin and upgraded the core version

unandyala:


monkey-jeff (dismissed):

.depcheckrc.yml
```diff
@@ -0,0 +1,3 @@
+ignores:
+  - typedoc-plugin-external-module-map
```
wjhsf, line 2:
```suggestion
# Include an explanation for the dependency is listed!
  - typedoc-plugin-external-module-map # Used by typedoc in `npm run doc`
```
I've noticed that we have a tendency to include things that don't need to be (which sometimes causes errors), so including explanations could help keep things clean and running.

unandyala: Agree. I will do in a separate PR since this is already merged

PR #267, opened 2020-09-21
Examples to retrieve auth token

unandyala:
Examples to retrieve 
1. authorization token for guest and registered customer
2. refresh token

### DIFF
--- a/examples/01-guest-shopper-auth-token.ts
+++ b/examples/01-guest-shopper-auth-token.ts
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2020, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * Get authorization token for a guest user
+ * Usage: ts-node examples/01-guest-shopper-auth-token.ts
+ * Note: Replace configuration parameters before running
+ */
+import { helpers, ClientConfig } from "../dist";
+
+//client configuration parameters
+const clientConfig: ClientConfig = {
+  parameters: {
+    clientId: "your-client-id",
+    organizationId: "your-org-id",
+    shortCode: "your-short-code",
+    siteId: "your-site-id",
+  },
+};
+
+/**
+ * Invoke helper function to retrieve the authorization token for the guest user
+ * Doc: https://salesforcecommercecloud.github.io/commerce-sdk/modules/_helpers_.html#getshoppertoken
+ */
+helpers
+  .getShopperToken(clientConfig, { type: "guest" })
+  .then((shopperToken) => {
+    console.log("Authorization token: ", shopperToken.getAuthToken());
+    console.log("Customer Info: ", shopperToken.getCustomerInfo());
+  })
+  .catch((error) => {
+    console.log(`Error fetching token: ${error}`);
+  });
--- a/examples/01-guest-shopper-auth-token.ts
+++ b/examples/01-guest-shopper-auth-token.ts
@@ -29,8 +29,14 @@ const clientConfig: ClientConfig = {
 helpers
   .getShopperToken(clientConfig, { type: "guest" })
   .then((shopperToken) => {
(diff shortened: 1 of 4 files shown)


examples/02-registered-shopper-auth-token.ts
```diff
@@ -0,0 +1,75 @@
...
+  ShopperToken<ShopperCustomers.Customer>
+> {
+  const credentials = `${username}:${password}`;
+  const buff = Buffer.from(credentials);
+  const base64data = buff.toString("base64");
+  const headers = { Authorization: `Basic ${base64data}` };
+
+  const client = new ShopperCustomers(clientConfig);
+
+  const response: Response = await client.authorizeCustomer(
```
wjhsf, line 51:
```suggestion
  const response = await client.authorizeCustomer(
```
`Response` is not necessary, and it's not a valid type anyway.

examples/01-guest-shopper-auth-token.ts
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+
+/**
+ * Get authorization token for a guest user
+ * Usage: ts-node examples/01-guest-shopper-auth-token.ts
+ * Note: Replace configuration parameters before running
+ */
+import { helpers, ClientConfig } from "../dist";
```
wjhsf, line 13:
```suggestion
import { helpers, ClientConfig } from "@commerce-sdk";
```
Importing from `dist` requires being in this repo, installing everything, and running the build process.

Importing from `commerce-sdk` works anywhere and only requires installing `commerce-sdk`.

wjhsf (changes requested):
Requesting changes specifically for the potentially breakable code in example 3.

examples/02-registered-shopper-auth-token.ts
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Get token for the registered customer
+ *
+ * @returns authorization token
+ */
+async function getRegisteredShopperToken(): Promise<
+  ShopperToken<ShopperCustomers.Customer>
+> {
+  const credentials = `${username}:${password}`;
```
alagopus, line 44: I believe username and password need to be treated with `encodeURIComponent` before being encoded with base64 again. It sounds redundant, but that's what it says here: https://tools.ietf.org/html/rfc6749#page-16

examples/01-guest-shopper-auth-token.ts
unandyala: Referring/Importing the same repo gives an error.  Since examples are in the repo, I think this is easier  to run

examples/02-registered-shopper-auth-token.ts
```diff
@@ -0,0 +1,75 @@
...
+ * Note: Replace configuration parameters and user credentials before running
+ */
+import { ClientConfig, Customer } from "../dist";
+import {
+  getObjectFromResponse,
+  ResponseError,
+  ShopperToken,
+  stripBearer,
+} from "@commerce-apps/core";
+import ShopperCustomers = Customer.ShopperCustomers;
```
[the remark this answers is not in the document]
unandyala: It just creates an alias - https://www.typescriptlang.org/docs/handbook/namespaces.html#aliases

I believe the const works only for type aliases

wjhsf (approved):

examples/02-registered-shopper-auth-token.ts
unandyala: I have checked with our API team and confirmed that this API is not designed per OAuth2.0 standards. Auth header should be ‘Basic ’ + base64 (username + ‘:’ + password).

bredmond-sf (approved):

gkhunger-sf (approved):

examples/02-registered-shopper-auth-token.ts
alagopus: Interesting. During the online workshop we discovered that passwords containing % don't work. Maybe that's been changed. Or I am confusing the endpoint with a different full Oauth2 thing in the SFCC arena.

PR #271, opened 2020-10-13
Update dtoPartial template to handle array types

gkhunger-sf:
Build is currently failing as it depends on https://github.com/SalesforceCommerceCloud/raml-toolkit/pull/128

### DIFF
--- a/templates/ClientInstance.ts.hbs
+++ b/templates/ClientInstance.ts.hbs
@@ -7,7 +7,7 @@ import { OperationOptions } from "retry";
 export namespace {{name.upperCamelCase}} {
   {{#each dataTypes}}
       {{#if (isTypeDefinition .)}}
-      export type {{getValue name}} = {{> dtoPartial typeDto=.}}
+export type {{getValue name}} = {{> dtoPartial typeDto=.}}
       {{/if}}
   {{/each}}
 }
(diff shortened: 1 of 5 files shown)


templates/ClientInstance.ts.hbs
```diff
@@ -6,8 +6,8 @@ import { OperationOptions } from "retry";
 // eslint-disable-next-line @typescript-eslint/no-namespace
 export namespace {{name.upperCamelCase}} {
   {{#each dataTypes}}
-      {{#if (isTypeDefinition .)}}
-      export type {{getValue name}} = {{> dtoPartial typeDto=.}}
+      {{#if (or (isTypeDefinition this) (isArrayType this))}}
+export type {{getValue name}} = {{> dtoPartial typeDto=.}}
```
unandyala, line 10: whys is "dot" replaced with "this"?should we use "this" at line #10 as well to be consitent

gkhunger-sf: I think this is just the nicer way to doing it. But the effect is the same. I'll change it on line 10 too.

unandyala (approved):

bredmond-sf (approved):

monkey-jeff (approved):

PR #274, opened 2020-10-23
Invoke diff command after downloading the APIs

unandyala:


package.json
```diff
@@ -22,6 +22,7 @@
     "build": "npm run clean && npm run renderTemplates && tsc && eslint dist --ext .ts,.js --quiet --fix",
     "clean": "rm -rf renderedTemplates dist",
     "depcheck": "depcheck",
+    "diffApis": "raml-toolkit diff --dir updateApisTmp/oldApis apis -o updateApisTmp/ramlDiff.json",
```
shirishmaheshwari, line 25: Do we have any test cases to cover this scenario?

unandyala: The diff functionality has tests  written in raml-toolkit

bredmond-sf (approved):

package.json
```diff
@@ -33,7 +34,7 @@
     "pretest": "npm run lint && npm run depcheck",
     "test": "nyc mocha \"src/**/*.test.ts\"",
     "test:ci": "npm test -- --reporter=xunit --reporter-options output=./reports/generator.xml",
-    "updateApis": "ts-node src/updateApis.ts"
+    "updateApis": "npm run backupApis && ts-node src/updateApis.ts && npm run diffApis"
```
kalagarraj, line 37: Is there a command that updates the APIs without diffs? if not would it make sense to add one

kalagarraj (approved):
approved with a minor suggestion

package.json
unandyala: do we have a use case for just download? We almost always want to diff. 
We can add a command to do backup and then update APIs (without diffs) if it is useful. This way we will able to run diffs if required after the download.

shirishmaheshwari (approved):

PR #280, opened 2020-12-02
Add error doc

bredmond-sf:
Added details about Errors to the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,29 @@ helpers.getShopperToken(config, { type: "guest" }).then(async (token) => {
 
 ### Error Handling
 
-SDK methods return an appropriate object by default when the API call returns a successful response. The object is built from the body of the response. If the API response is not successful, an Error is thrown. The error message is set to the status code plus the status text. The Error object includes a custom 'response' attribute with the entire Response object for inspection. To access the body of the response when an error occurs, see sample code above.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -105,6 +107,12 @@ helpers.getShopperToken(config, { type: "guest" }).then(async (token) => {
 });
 ```
 
+### Error Handling
+
+SDK methods return an appropriate object by default when the API call returns a successful response. The object is built from the body of the response. If the API response is not successful, an Error is thrown. The error message is set to the status code plus the status text. The Error object includes a custom 'response' attribute with the entire Response object for inspection. To access the body of the response when an error occurs, see sample code above.
```
jbegleiter-sf, line 112: Can we add one concrete example to this to make it clear why someone would want to do this, e.g. in the case where you want to see the server response for why a request is bad (400)?

bredmond-sf: Added an example, let me know what you think

monkey-jeff (approved):

jbegleiter-sf (approved):

unandyala (approved):

tobiaslohr (commented):
Shouldn't we allow logging of the `e.response.text()` through verbose logging out of the box? This may help with troubleshooting without actually touching the client implementation itself.

jbegleiter-sf:
@tobiaslohr the problem with doing this is that it interprets/expends the buffer represented by the response text. It has the potential to change the behavior the consumer sees if they enable verbose mode, which is bad behavior. As a result we decided intentionally to make it explicit here. There are alternatives but they would represent a feature request/enhancement as opposed to adding this documentation.

PR #293, opened 2021-02-09
Add CCDC object ids of inventory and slas admin apis

gkhunger-sf:
Add inventory and slas admin apis to the object map so that docs can be pulled from CCDC
Also, move cc api families list to config.ts

### DIFF
--- a/src/updateApis.ts
+++ b/src/updateApis.ts
@@ -8,16 +8,12 @@
 import path from "path";
 import fs from "fs-extra";
 import { updateApis } from "./lib/utils";
-import { API_FAMILIES } from "./lib/config";
-
-const PRODUCTION_API_PATH = `${__dirname}/../apis`;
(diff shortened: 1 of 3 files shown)


unandyala (dismissed):

src/updateApis.ts
```diff
@@ -8,15 +8,7 @@
...
-
-const API_FAMILIES = [
-  "pricing",
-  "customer",
-  "checkout",
-  "search",
-  "product",
-  "cdn",
-];
+import { API_FAMILIES } from "./lib/config";
```
wjhsf, line 11: Should the other constants in this file be moved too?

gkhunger-sf: fine 🙄

wjhsf (approved):

unandyala (approved):

PR #311, opened 2021-03-16
Add snyk test to package.json and other changes snyk suggested

gkhunger-sf:
Just ran snyk wizard

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,6 +1,11 @@
 # Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
-version: v1.14.1
-ignore: {}
+version: v1.19.0
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+ignore:
+  SNYK-JS-SSRI-1085630:
+    - '@commerce-apps/core > ssri':
+        reason: No known remediation available
+        expires: '2021-04-15T21:26:18.209Z'
 # patches apply the minimum changes required to fix a vulnerability
 patch:
   SNYK-JS-LODASH-567746:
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -32,7 +32,7 @@
     "prepare": "snyk protect",
     "renderTemplates": "ts-node src/generate.ts",
     "pretest": "npm run lint && npm run depcheck",
-    "test": "nyc mocha \"src/**/*.test.ts\"",
+    "test": "snyk test && nyc mocha \"src/**/*.test.ts\"",
```
unandyala, line 35: why are we adding synk test to the test command?

.snyk
```diff
@@ -1,6 +1,11 @@
 # Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
-version: v1.14.1
-ignore: {}
+version: v1.19.0
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+ignore:
+  SNYK-JS-SSRI-1085630:
```
wjhsf, line 5: This change is outdated. A remediation is now available. https://snyk.io/vuln/SNYK-JS-SSRI-1085630#remediation

package.json
wjhsf: The result of `snyk test` will change over time, so test results will not be reproducible. I don't think that we should do this. Also because we already have the same functionality from our Snyk monitoring integrations. (I think. Snyk docs aren't _super_ clear.)

If we do want `snyk test` to be run, it should be as a `postinstall` script, not a `test` script. Like `npm audit` is.

wjhsf (changes requested):

gkhunger-sf:
snyk test doesn't work without auth. Let's talk about it and do it properly if we want to do it.

PR #322, opened 2021-04-15
Add support for RAML traits

wjhsf:
Recent changes to APIs require that we now include type definitions for RAML traits.

There are a handful of APIs that use a trait named `offset-paginated`, which is not a valid TypeScript identifier (and is noncompliant with the API Standards), so I also added a helper to explicitly ignore that trait. None of our type definitions currently use the trait, so it can safely be excluded. (An equivalent `OffsetPaginated` will be added to bring the APIs into compliance, which we will be able to use if it ever becomes necessary.)

**Note:** Builds will fail on this PR until https://github.com/SalesforceCommerceCloud/raml-toolkit/pull/176 is merged and a new version of the RAML toolkit is released.


src/lib/templateHelpers.ts
```diff
@@ -131,3 +132,14 @@ export const isCommonQueryParameter = (property: string): boolean =>
...
+
+/**
+ * Checks whether a trait name is allowed to be used in the API.
+ * Currently only disallows "offset-paginated", as "OffsetPaginated" is also
+ * available, and is preferred.
+ *
+ * @param trait - Trait to check
+ * @returns true unless the trait's name is "offset-paginated"
+ */
+export const isAllowedTrait = (trait: amf.model.domain.Trait): boolean =>
```
gkhunger-sf, line 144: Why are we hardcoding it instead of checking whether a trait is UpperCamelCase or not?

wjhsf: According to the API standards, traits are supposed to be camel case. Currently, `offset-paginated` is the only trait that is noncompliant. Theoretically, no more noncompliant traits will be added. However, if any are, we should treat that as an error and fix it, rather than silently ignore it (which could cause further issues).

gkhunger-sf: We can log the non-compliant traits and try to get them fixed without breaking SDK generation. This would mean that SDK generation is broken until we either get the trait fixed (which could take time) or add another exception in the code.

unandyala, line 144: Can we add a detailed note that RAMLs are planned to update to use standards and this will remain mainly for backward compatibility

unandyala (approved):

src/lib/templateHelpers.ts
wjhsf: done

echessman (approved):

gkhunger-sf (approved):

unandyala (approved):

PR #332, opened 2021-07-20
Add user agent header

wjhsf:
This will fail until https://github.com/SalesforceCommerceCloud/raml-toolkit/pull/191 is merged and released.

This implementation won't allow users to specify a custom `user-agent` header, but they can override the default by using a different case (e.g. `User-Agent`). I don't think this is a significant concern, so I didn't address it, but it's there.


kalagarraj (changes requested):
headers are case insensitive. Pls change the implementation to accept custom user agents in addition to adding the commerce sdk header

wjhsf:
We don't expect users to provide their own user agent (it doesn't change the behavior of any of the APIs), so I don't think we need to add any logic for an edge case we don't expect to encounter.

bredmond-sf (approved):

kalagarraj:
> We don't expect users to provide their own user agent (it doesn't change the behavior of any of the APIs), so I don't think we need to add any logic for an edge case we don't expect to encounter.

Actually It is not about an edge case, we must try to follow the RFC's as much as possible. The RFC for HTTP header states that user agents are case insensitive and can have multiple values.

wjhsf:
We already treat things in a case insensitive manner, but we specifically [don't allow multiple values](https://github.com/SalesforceCommerceCloud/commerce-sdk-core/blob/ace1e8a190598e1eff90225ed1ba6f86a0703bac/test/staticClient/headers.test.ts#L202).

testIntegration/tests/userAgent.test.ts
```diff
@@ -0,0 +1,73 @@
...
+"use strict";
+import fs from "fs-extra";
+import nock from "nock";
+import { expect } from "chai";
+import { ShopperCustomers } from "commerce-sdk/dist/customer/customer";
+
+// Helper to make the user agent being tested a bit more obvious
+async function mockRequestWithUserAgent(userAgent: string): Promise<void> {
+  nock("http://somewhere")
+    .matchHeader("user-agent", userAgent)
```
kalagarraj, line 16: add tests for client specific user agents are not overwritten

wjhsf: Added.

templates/operations.ts.hbs
```diff
@@ -145,6 +145,11 @@
         {{/if}}
       {{/each}}
 
+      const headers = (options && options.headers) ? { ...options.headers } : {};
```
unandyala, line 148: Can this be a helper function so that we can test?

wjhsf: This is just a getter with a default value. Not much to really test here.

kalagarraj (approved):

unandyala (dismissed):

echessman (approved):

gkhunger-sf (approved):

echessman (approved):

PR #356, opened 2022-02-28
W-10495391: Add SLAS helper

joeluong-sfcc:
This PR adds SLAS helper functions to help users easily onboard SLAS. This PR also consumes an update from commerce-sdk-core that allows `fetchOptions` to be exposed in the client config.


tarcang (approved):

templates/index.ts.hbs
```diff
@@ -7,10 +7,14 @@ export {
 {{/each}}
 }
 
+{{!-- TODO: consider resolving the helpers as one --}}
+{{!-- NOTE: helpers is deprecated since it uses shopperCustomer --}}
```
wjhsf, line 11: Suggestion:
1. Create `src/static`, like the iso sdk has. When rendering the templates, just copy the static files to the output directory.
2. Create a `src/static/helpers` directory.
3. `src/static/helpers/slas.ts` should have the SLAS helper.
4. `src/static/helpers/shopperCustomers.ts` or whatever should have the old helper. (It never changes, it doesn't need to be generated.)
4.b. Clean up the [metadata](https://github.com/SalesforceCommerceCloud/commerce-sdk/blob/master/apis/.metadata.json) that we don't need any more. (Side note, we can probably also clean up the `enforceCcdcId` stuff now, too.)
5. `src/static/helpers/index.ts` should `export * from "./whatever.ts"`
6. `src/index.ts` could either `export * as helpers from "./helpers"` or `export * from "./helpers"`. Not sure which is better.

Also, wrap the deprecated helper with [`util.deprecate`](https://nodejs.org/docs/latest-v12.x/api/util.html#util_util_deprecate_fn_msg_code).

src/slasHelper/slasHelper.ts
```diff
@@ -0,0 +1,263 @@
...
+      redirect_uri: parameters.redirectURI,
+      response_type: "code",
+      ...(parameters.usid && { usid: parameters.usid }),
+    },
+  };
+
+  const response = await slasClientCopy.authorizeCustomer(options, true);
+  const redirectUrl = response.headers?.get("location") || response.url;
+
+  // url is a read only property we unfortunately cannot mock out using nock
```
monkey-jeff, line 118: But aren't we passing it as a function parameter? Can't we just pass in a falsy value?

src/slasHelper/slasHelper.ts
```diff
@@ -0,0 +1,263 @@
...
+/**
+ * Encodes a code verifier to a code challenge to send to the authorization endpoint
+ * @param codeVerifier - random string to use as a code verifier
+ * @returns code challenge
+ */
+export const generateCodeChallenge = async (
+  codeVerifier: string
+): Promise<string> => {
+  const urlSafe = (input: string) =>
+    input.replace(/\+/g, "-").replace(/\//g, "_").replace(/=/g, "");
```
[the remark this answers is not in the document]
wjhsf: [It's a standard variant of base 64.](https://www.base64url.com/)

src/slasHelper/slasHelper.ts
```diff
@@ -0,0 +1,263 @@
...
+
+  const crypto = await import("crypto");
+  const challenge = urlSafe(
+    crypto.default.createHash("sha256").update(codeVerifier).digest("base64")
+  );
+
+  /* istanbul ignore next */
```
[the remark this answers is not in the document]
wjhsf: This was copied from the isomorphic SDK, where limitations of the testing environment prevented the block from being reached.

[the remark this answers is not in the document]
joeluong-sfcc: Apologies! I copied and pasted directly from the isomorphic SDK to have a baseline to work with. I did a quick pass through to remove all client side code but I forgot some of these comments we put in for testing. Thanks for catching this!

templates/index.ts.hbs
monkey-jeff: This cleans up what I was saying below that these things shouldn't be split between and static and template.  Well articulated well.

johnboxall:
@joeluong-sfcc – server-side (like in Commerce SDK), it is common to use a SLAS *private* client.

In the PKCE flow for a SLAS private client, you must provide `Authorization: Bearer B64($client_id:$client_secret)` to  `getAccessToken`.

A small diff fix might be allowing folks to pass `headers` to this method – alternatively you could optionally accept "credentials" which would contain a SLAS API Client ID & Secret.

Screen Shot 2022-03-08 at 10 43 27 AM (1)

joeluong-sfcc:
> @joeluong-sfcc – server-side (like in Commerce SDK), it is common to use a SLAS _private_ client.
> 
> In the PKCE flow for a SLAS private client, you must provide `Authorization: Bearer B64($client_id:$client_secret)` to `getAccessToken`.
> 

Hey @johnboxall, thanks for pointing this out! I'm still quite unfamiliar with OAuth v2.0 so the code example and picture really helped.


> Alternatively, is this already possible via the `slasClient` param? Something like this:
> 
> ```js
> import { helpers, Customers } from "commerce-sdk"
> 
> const client = new Customers.ShopperLogin({
>     headers: { 
>         Authorization: "Bearer " + Buffer.from("id:secret").toString("base64")
>     }
> })
> await helpers.loginRegisteredUserB2C(client)
> ```
> 

I believe this works!

> As part of this change, can we consider updating `README.md` to highlight using these methods as they are now the entry way to SCAPI?

Yep! I'm planning to update the README to with examples on how to use the helpers once I start tying things up.

joeluong-sfcc:
PR is currently blocked until https://github.com/SalesforceCommerceCloud/commerce-sdk-core/pull/93 is merged and released

src/static/helperTemplates/slas.ts.hbs
```diff
@@ -0,0 +1,272 @@
...
+      response_type: "code",
+      ...(parameters.usid && { usid: parameters.usid }),
+    },
+  };
+
+  const response = await slasClientCopy.authorizeCustomer(options, true);
+  const redirectUrl = response.headers?.get("location") || response.url;
+
+  if (!redirectUrl) {
+    throw new Error("Authorization failed");
```
johnboxall, line 115: Hey @joeluong-sfcc – on `commerce-sdk-isomorphic`, one thing thats blocking us from adopting the helpers is lack of access to the underlying errors.

This first call may fail for a variety of reasons:

1. SLAS miss configured (bad tenant, bad client, bad redirect URI)
2. Bad code challenge (repeated, invalid)
3. No network, transient network error etc.

I believe the SDK already has a pattern of attaching the `response` object to the error so that callers can handle it and display an appropriate error message to the user.

Can we use it here as well? <3 <3 <3

wjhsf: `ResponseError` [exists in the core package](https://github.com/SalesforceCommerceCloud/commerce-sdk-core/blob/096f14d1300b3a26cce2ce8abd62ac5080d65035/src/base/staticClient.ts#L51), and can be used here.

src/static/helpers/slas.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  if (!redirectUrl) {
+    throw new Error("Authorization failed");
+  }
+
+  return { url: redirectUrl, ...getCodeAndUsidFromUrl(redirectUrl) };
+}
+
+/**
+ * A single function to execute the ShopperLogin Public Client Guest Login with proof key for code exchange flow as described in the [API documentation](https://developer.salesforce.com/docs/commerce/commerce-api/references?meta=shopper-login:Summary).
```
johnboxall, line 127: Nit: If we're asking folks provide a client secret, then must use a SLAS _Private Client_ w/ PKCE.

src/static/helpers/slas.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  // Create a copy in order to modify behavior of the fetch call
+  // We do not want to redirect so manually control redirect behavior
+  const slasClientCopy = new ShopperLogin(slasClient.clientConfig);
+  slasClientCopy.clientConfig.fetchOptions = {
+    ...slasClient.clientConfig.fetchOptions,
+    redirect: "manual",
+  };
+
+  const response = await slasClientCopy.authenticateCustomer(
```
johnboxall, line 210: As above, there are many reasons why this call may fail. We'll want to inspect `response` for errors (bad status codes / lack of a redirect URL) and throw with enough information for the caller to resolve.

src/static/helpers/slas.ts
johnboxall: Hey @joeluong-sfcc, the comment of this function says:

> A single function to execute the ShopperLogin Public Client Guest Login with proof key for code exchange flow as described in the API documentation.

But the implementation is a private client flow using client credentials. _NOT a public client with PKCE_.

I'm saying the function comment should match what the function does.

unandyala (dismissed):

src/static/helperTemplates/slas.ts.hbs
```diff
@@ -0,0 +1,423 @@
+/*
```
unandyala, line 1: This looks good but I feel this should be a helper instead of a template. Helpee can accept the instance of the shopperLogin class

unandyala (approved):

.eslintignore
```diff
@@ -1,2 +1,2 @@
 renderedTemplates
-
+dist/**/*.ts
```
unandyala, line 2: Is this to ignore DTS file? Was it linting the DTS files successfully before these changes?

unandyala (dismissed):

.eslintignore
wjhsf: If linting is reporting an error, then you should fix the error, rather than disabling linting. In this case, the comment blocks are not the expected format, so they should be changed. (They are jsdoc, not tsdoc. There are minor but annoying differences.)

src/static/helpers/slasClient.ts
```diff
@@ -0,0 +1,107 @@
...
+export type TokenResponse = {
+  access_token: string;
+  id_token: string;
+  refresh_token: string;
+  expires_in: number;
+  token_type: string;
+  usid: string;
+  customer_id: string;
+  enc_user_id: string;
+} & { [key: string]: any };
```
wjhsf, line 42: Aren't these types defined in the generated `shopperLogin.ts`? You could import them, but you'd also have to be careful about circular dependencies.

.npmrc
```diff
@@ -0,0 +1 @@
+registry=https://registry.npmjs.org
```
wjhsf, line 1: Out of curiosity, what is this for?

.depcheckrc.yml
```diff
@@ -2,3 +2,4 @@ ignores:
   # Include a reason why each package is listed
   - typedoc-plugin-external-module-map # Used by typedoc in `npm run doc`
   - typedoc-plugin-nojekyll # Used by typedoc in `npm run doc`
+  - node-fetch # We only use @types/node-fetch, not node-fetch itself
```
wjhsf, line 5: If you're not using `node-fetch`, you may want to use the types for the fetch package you actually use (iirc `make-fetch-happen`, but maybe `minipass-fetch`).

examples/README.md
```diff
@@ -1,6 +1,6 @@
 # Commerce SDK Examples
 
-This directory contains example usage of the Commerce SDK. Each example contains valid credentials for a public demo instance. The examples can be executed successfully as-is, or you can edit them to provide your own credentials.
+This directory contains example usage of the Commerce SDK. Each example contains valid credentials for a public demo instance. Examples 01-03 can be executed successfully as-is (or you can edit them to provide your own credentials). Examples 04-06 require additional editing (marked by TODO).
```
wjhsf, line 3: Maybe mark as `EDIT HERE` or something that's more clear it's for the user to do, not an incomplete example?

src/static/helpers/slas.ts
```diff
@@ -0,0 +1,409 @@
...
+  slasClient: ISlasClient,
+  credentials: {
+    username: string;
+    password: string;
+    clientSecret: string;
+  },
+  parameters: {
+    redirectURI: string;
+    usid?: string;
+  }
```
wjhsf, line 209: Not 100% on this, but iirc defining object types within function signatures like this can cause issues with the generated docs. But it may have been _nested_ objects (e.g. `options: { credentials: { ... }, parameters: { ... } }`), and it's been more than a year, so maybe the bug's been patched.

wjhsf (commented):
Mostly the linting issue...

.npmrc
joeluong-sfcc: Only node 16 is broken on the CI and its due to a nexus connectivity issue. We're holding off merging until we fix the CI, this was a hopeful shot in the dark that this would get it working. It did not work 😭. I plan on reverting this change when we get closer to merging

.depcheckrc.yml
joeluong-sfcc: So currently I only have `"@types/node-fetch": "^2.5.12",` as a dev dependency and not `node-fetch` since I only want the type definitions for fetch options and not the functionality of `node-fetch`. Would it be best to use `make-fetch-happen`/`minipass-fetch` instead or just the types (`@types/make-fetch-happen)?

src/static/helpers/slasClient.ts
joeluong-sfcc: `If linting is reporting an error, then you should fix the error, rather than disabling linting. In this case, the comment blocks are not the expected format, so they should be changed. (They are jsdoc, not tsdoc. There are minor but annoying differences.)`

Responding to your linting here cause I couldn't find an easy way to reply that's my b:

So usually I try to fix linting errors as they arise rather than disabling linting. Here's an example of the way it's currently documented:

```javascript
/**
 * A single function to execute the ShopperLogin Public Client Guest Login with proof key for code exchange flow as described in the [API documentation](https://developer.salesforce.com/docs/commerce/commerce-api/guide/slas-public-client.html).
 * @param slasClient a configured instance of the ShopperLogin SDK client.
 * @param parameters - parameters to pass in the API calls.
 * @param parameters.redirectURI - Per OAuth standard, a valid app route. Must be listed in your SLAS configuration. On server, this will not be actually called
 * @param parameters.usid? - Unique Shopper Identifier to enable personalization.
 * @returns TokenResponse
 */
export async function loginGuestUser(
  slasClient: ISlasClient,
  parameters: {
    redirectURI: string;
    usid?: string;
  }
):
```

The linter is complaining about the dot notation, which I believe is the most intuitive way to write and understand the documentation. Also from what I've seen/researched so far (including JSDoc), this style of is pretty standard on documenting parameters with properties. I'd also like to keep the documentation changes as close as to the `isomorphic sdk` as well just for consistency. I'm totally cool with changing the documentation if you have a convincing argument for the other side though 👍

wjhsf: Do what you want, but if there's just one rule in one file that's causing trouble and you don't want to fix it, then just disable the one rule in the one file, not every rule in every file. 😉

joeluong-sfcc: Totally fair point. I figured since `renderedTemplates` was in `.eslintignore` I could toss `dist` as well since it's also dynamically generated.

gkhunger-sf (approved):

PR #376, opened 2022-11-29
chore: switch circle CI with GH actions

joeluong-sfcc:
This PR replaces circle CI functionality with GH actions. This PR also updates the default branch of `master` to `main`.

Workflows added:
- `test.yml`: runs unit tests and integration tests against node versions 10, 12, 14, 16
- `publish.yml`: publishes to NPM when a github release is created
    - can also manually trigger if needed
    - NPM token is stored as secret and needs to be rotated every year


monkey-jeff:
When we first created this repo we went with Circle CI because security said that the use of GH Actions and Travis CI were not approved for use, but Circle CI was approved at the time.

I am curious as to if this has changed.

joeluong-sfcc:
@monkey-jeff So from my understanding, GH actions was originally not allowed but now is security approved. We're moving from circle CI to GH actions because Circle CI doesn't allow static IPs for whitelists (at least for the plan salesforce/commerce cloud is on)

.github/workflows/test.yml
```diff
@@ -0,0 +1,36 @@
...
+          cache: npm
+      - name: Cache node modules
+        id: cache-nodemodules
+        uses: actions/cache@v3
+        env:
+          cache-name: cache-node-modules
+        with:
+          path: "**/node_modules"
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/yarn.lock') }}
+      - run: npm install --legacy-peer-deps
```
unandyala, line 29: why is this step required?

joeluong-sfcc: The `--legacy-peer-deps` flag was added because there seems to be a peer dependency conflict due to eslint that causes `npm install` to fail. Here's an example of `npm install` failing:

https://github.com/SalesforceCommerceCloud/commerce-sdk/actions/runs/3576294833/jobs/6013930295

unandyala: are the conflicting versions are from this repo. can we resolve the conflict by upgrading the version?

joeluong-sfcc: it might, let me go ahead and try changing around versioning. If it seems like this requires more work than a few minor tweaks, I'll create a follow up story

unandyala (approved):

p-biyani (approved):

PR #381, opened 2022-12-27
fix: update admin API examples

joeluong-sfcc:
@W-11495342@


tarcang (commented):
Looks very good. After adding some tests, it should be good 👌

tarcang (approved):

unandyala (approved):

PR #395, opened 2023-10-24
Release/20231024

adamraya:
#### New APIs

* *Shopper Experience* has been added to the SDK.


src/lib/config.ts
```diff
@@ -18,7 +18,7 @@ export const API_FAMILIES = [
   "checkout",
   "customer",
   "discovery",
-  "inventory",
```
wjhsf, line 21: `inventory` has been in the API list for years, but we've [never actually released it](https://www.npmjs.com/package/commerce-sdk?activeTab=code), because it's always been broken (and nobody's asked for it). It's just a pain to deal with, so we decided to remove it.

wjhsf (approved):

joeluong-sfcc (approved):

PR #396, opened 2024-01-18
Feat: Add support for custom query params @W-14391914@

joeluong-sfcc:
This PR adds support for custom query parameters as SCAPI now supports SCAPI hooks: https://developer.salesforce.com/docs/commerce/commerce-api/references/about-commerce-api/about.html#08302023

Query params that begin with `c_` are considered custom query parameters and are passed onto the underlying SCAPI call.

This PR also:
- adds filtering for invalid query parameters
- upgrades `typescript` and `ts-node` packages and fixes respective typescript errors due to upgrade
- updates CI to include node 18 and 20

To test this PR, you can build a local version of the `commerce-sdk` and use that as a package for manual testing:

1. `git clone -b feat/support-custom-params git@github.com:SalesforceCommerceCloud/commerce-sdk.git`
2. `cd commerce-sdk`
3. `npm install --legacy-peer-deps`
4. Add a `console.log` to `templates/operations.ts.hbs` to see query params in console output when testing
```javascript
...
      console.log('queryParams: ', queryParams)
      return StaticClient.{{method}}({
        client: this,
        rawResponse: rawResponse,
...
```
5. `npm run build`
6. Create new testing directory:
```bash
mkdir testCustomParams
cd testCustomParams
npm init -y
touch index.js
```
7. modify package.json in testCustomParams directory to point `commerce-sdk` to local directory in dependencies section and add "type": "module"
```json
  "type": "module",
  "dependencies": {
    "commerce-sdk": "<insert_path_to_local_commerce-sdk>"
  }
```
8. `npm install`
9. Add test code to `index.js` and replace client creds
```javascript
import * as CommerceSdk from "commerce-sdk";
const { slasHelpers, Customer, Product } = CommerceSdk;

// demo client credentials, if you have access to your own please replace them below.
const CLIENT_ID = "<CLIENT_ID>";
const ORG_ID = "<ORG_ID>";
const SHORT_CODE = "<SHORT_CODE>";
const SITE_ID = "<SITE_ID>";

// must be registered in SLAS. On server, redirectURI is never called
const redirectURI = "http://localhost:3000/callback";

// client configuration parameters
const clientConfig = {
  parameters: {
    clientId: CLIENT_ID,
    organizationId: ORG_ID,
    shortCode: SHORT_CODE,
    siteId: SITE_ID,
  },
};

const slasClient = new Customer.ShopperLogin(clientConfig);

// GUEST LOGIN
const guestTokenResponse = await slasHelpers
  .loginGuestUser(slasClient, { redirectURI })
  .then((guestTokenResponse) => {
    console.log("Guest Token Response: ", guestTokenResponse);
    return guestTokenResponse;
  })
  .catch((error) =>
    console.log("Error fetching token for guest login: ", error)
  );

const shopperProducts = new Product.ShopperProducts({
    ...clientConfig,
    headers: {authorization: `Bearer ${guestTokenResponse.access_token}`}
});

const productsResult = await shopperProducts.getProducts({
    parameters: {
        ids: "25720044M,25686395M",
        expand: ["promotions", "images", "prices"],
        invalid_query: 'invalidQueryParameter',
        c_customQuery: 'validQueryParameter'
    }
})

console.log("productsResult: ", productsResult)
```
10. `node index.js > out.txt`
11. Observe out.txt and that the custom query params are included and invalid query params are excluded from the call

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,18 +18,18 @@ jobs:
         with:
           node-version: ${{ matrix.node_version }}
           cache: npm
-      - name: Cache node modules
-        id: cache-nodemodules
-        uses: actions/cache@v3
-        env:
-          cache-name: cache-node-modules
-        with:
-          path: "**/node_modules"
-          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/yarn.lock') }}
+      # - name: Cache node modules
+      #   id: cache-nodemodules
+      #   uses: actions/cache@v3
+      #   env:
+      #     cache-name: cache-node-modules
+      #   with:
+      #     path: "**/node_modules"
+      #     key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/yarn.lock') }}
       # - run: npm install snyk --legacy-peer-deps
       # - run: npm run snyk:auth ${{ secrets.SNYK_TOKEN }}
       - run: npm install --legacy-peer-deps
-        if: ${{ steps.cache-nodemodules.outputs.cache-hit != 'true' }}
+        # if: ${{ steps.cache-nodemodules.outputs.cache-hit != 'true' }}
       - run: npm run build
       - name: Unit tests
         run: npm test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,7 +29,7 @@ jobs:
       - run: npm install snyk --legacy-peer-deps
       - run: npm run snyk:auth ${{ secrets.SNYK_TOKEN }}
       - run: npm install --legacy-peer-deps
-        # if: ${{ steps.cache-nodemodules.outputs.cache-hit != 'true' }}
+        if: ${{ steps.cache-nodemodules.outputs.cache-hit != 'true' }}
       - run: npm run build
       - name: Unit tests
         run: npm test
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,8 +26,8 @@ jobs:
         with:
           path: "**/node_modules"
           key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/yarn.lock') }}
-      - run: npm install snyk --legacy-peer-deps
-      - run: npm run snyk:auth ${{ secrets.SNYK_TOKEN }}
+      # - run: npm install snyk --legacy-peer-deps
+      # - run: npm run snyk:auth ${{ secrets.SNYK_TOKEN }}
       - run: npm install --legacy-peer-deps
         if: ${{ steps.cache-nodemodules.outputs.cache-hit != 'true' }}
       - run: npm run build
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
   linux-tests:
     strategy:
       matrix:
-        node: [10, 12, 14, 16, 18, 20]
+        node: [16, 18, 20]
       fail-fast: false
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,18 +18,18 @@ jobs:
         with:
           node-version: ${{ matrix.node_version }}
           cache: npm
-      # - name: Cache node modules
-      #   id: cache-nodemodules
-      #   uses: actions/cache@v3
-      #   env:
-      #     cache-name: cache-node-modules
-      #   with:
-      #     path: "**/node_modules"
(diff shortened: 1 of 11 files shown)


.github/workflows/test.yml
```diff
@@ -9,7 +9,7 @@ jobs:
   linux-tests:
     strategy:
       matrix:
-        node: [10, 12, 14, 16]
+        node: [10, 12, 14, 16, 18, 20]
```
alexvuong, line 12: Have we dropped support for 10-14?

package.json
```diff
@@ -79,11 +77,11 @@
     "sinon": "^9.2.4",
     "snyk": "^1.836.0",
     "sort-package-json": "^1.53.1",
-    "ts-node": "^9.1.1",
+    "ts-node": "^10.9.2",
```
alexvuong, line 80: Do these upgrade consider breaking change?

src/static/helpers/slas.ts
```diff
@@ -98,10 +98,10 @@ export async function authorize(
 
   const options = {
     parameters: {
-      client_id: slasClient.clientConfig.parameters.clientId,
+      client_id: slasClient.clientConfig.parameters?.clientId as string,
```
alexvuong, line 101: can parameters be undefined? 🤔

src/static/helpers/slasClient.ts
```diff
@@ -49,14 +49,15 @@ export interface ISlasClient {
     options: {
       parameters?: {
         organizationId?: string;
-      };
+      } & { [key in `c_${string}`]: any };
```
alexvuong, line 52: Can we use unknown over any type for this custom key value? 🤔 
```suggestion
      } & { [key in `c_${string}`]: unknown };
```

alexvuong:
Manual test as instructed in the PR. Confirming the code works as expected. I left a few non-blocking comments @joeluong-sfcc

.github/workflows/test.yml
joeluong-sfcc: I believe that the SDKs only need to support Node LTS, so technically yes, but thinking about leaving it in the CI as it doesn't hurt. If it starts causing issues we can be explicit in dropping support for older node versions

package.json
joeluong-sfcc: It shouldn't as `ts-node` and `typescript` are meant for development/building and not for the compiled SDK itself

src/static/helpers/slas.ts
joeluong-sfcc: I fixed this since typescript was complaining:
```
'slasClient.clientConfig.parameters' is possibly 'undefined'.ts(18048)
```

But yeah, I think it can be undefined,`parameters` is an optional property on the underlying type for `slasClient.clientConfig`:
```
export declare class ClientConfig {
    baseUri?: string;
    cacheManager?: ICacheManager;
    headers?: BasicHeaders;
    parameters?: CommonParameters;
    retrySettings?: OperationOptions;
    fetchOptions?: RequestInit;
}
```

src/static/helpers/slasClient.ts
joeluong-sfcc: Ah good callout, `unknown` is the the better type to use in this situation but unfortunately using unknown causes typescript errors :( 

```
Type 'Partial<{ [x: `c_${string}`]: unknown; }>' is not assignable to type 'QueryParameters'.
  '`c_${string}`' and 'string' index signatures are incompatible.
    Type 'unknown' is not assignable to type 'string | number | boolean | string[] | number[]'.
```

It's cause we send the query parameters to `commerce-sdk-core` which expects query parameters as typed:
`string | number | boolean | string[] | number[]`. So we can do:

```suggestion
      } & { [key in `c_${string}`]: string | number | boolean | string[] | number[] };
```

The only concern I have with this approach is that customers could theoretically pass other types to custom query parameters like `boolean []`. For that reason, I'm leaning towards using `any`, but I could be convinced otherwise

README.md
```diff
@@ -19,7 +17,7 @@ you import the type definitions directly.
 
 Download and install Node.js and npm [here](https://nodejs.org/en/download/).
 ​
-> **Note:** Only Node.js version 14 and 16 LTS are supported. Other versions can cause unexpected results. To use a different version of Node.js for other projects, you can manage multiple versions of Node.js with [nvm](https://github.com/nvm-sh/nvm). ​
+> **Note:** Only Node.js versions 14, 16, 18, and 20 are supported. Other versions can cause unexpected results. To use a different version of Node.js for other projects, you can manage multiple versions of Node.js with [nvm](https://github.com/nvm-sh/nvm). ​
```
vcua-mobify, line 20: Related to another comment above, since we only support Node LTS, should we remove 14 from this list?

alexvuong (approved):

vcua-mobify (approved):

alexvuong (approved):

PR #398, opened 2024-01-29
Release v2.13.0 @W-14879795@

joeluong-sfcc:
This PR prepares the release for v2.13.0. Changes include:
- Support for custom query parameters

For ease of review, the `docs/` directory is auto generated, and the `apis/` and `testIntegration/stagingApis` directory changes are pulled from anypoint exchange. There were some APIs that were added in `testIntegration/stagingApis` previously that we do not support in the SDK that are removed (baskets, data discovery ingestion, data discovery management)


README.md
```diff
@@ -213,6 +213,22 @@ client.authorizeCustomer({
...
+### Custom Query Parameters
+
+With the introduction of [hooks for Commerce APIs](https://developer.salesforce.com/docs/commerce/commerce-api/guide/extensibility_via_hooks.html), customers can pass custom query parameters through the SDK to be used in their custom hook. Custom query parameters must begin with `c_`:
+
+```javascript
+const searchResults = await searchClient.productSearch({
+  parameters: { 
+    q: "dress", 
+    limit: 5,
+    c_customQueryParam: '<your-custom-query-param>'
```
bendvc, line 225: Nit picky but if we want to be consistent with the naming with respect to the data object those names represent we might want something like this:

```
c_paramkey: '<param-value>'
```

the "custom" in `c_customQueryParam` is a little redundant because thats what "c" stands for, and "custom" in `'<your-custom-query-param>'` is a misnomer because all values are kind custom.

bendvc (approved):

PR #402, opened 2024-05-01
Feat: Add support for custom APIs @W-15585881@

joeluong-sfcc:
This PR adds support for custom APIs by exposing a helper function called `callCustomEndpoint` that allows users to call their custom API following common SDK patterns.

There is a companion PR in the `commerce-sdk-core` package that will need to be released first then consumed by this PR in order to pass the CI checks. The companion PR: https://github.com/SalesforceCommerceCloud/commerce-sdk-core/pull/110

Some features of the `callCustomEndpoint` helper:
1. Defaults `apiVersion` to `v1` if not provided
2. Defaults `content-type` header to `application/json` if not provided
3. options.body can be passed as an object and will be automatically formatted to match `Content-Type` through `commerce-sdk-core`

Below is a sample script I used for testing the `callCustomEndpoint` helper function. Reach out to me for instructions on how to run this with the proper credentials:

```javascript

import * as CommerceSdk from "commerce-sdk";
const { helpers } = CommerceSdk;

// client configuration parameters
const clientConfig = {
  parameters: {
    clientId: '<your_client_id>',
    siteId: '<your_site_id>',
    organizationId: '<your_org_id>',
    shortCode: '<your_short_code>',
  },
  // baseUri: 'https://{shortCode}.alternativeBaseUri.com/custom/{apiName}/{apiVersion}'
};

const access_token = '<insert_access_token_here>'

let response = await helpers.callCustomEndpoint({ options: {
    method: 'PUT',
    headers: {
        'Content-type': 'application/json',
        authorization: `Bearer ${access_token}`
    },
    customApiPathParameters: {
      apiVersion: 'v1',
      endpointPath: 'greeting',
      apiName: 'e2e-tests',
    },
    enableTransformBody: true,
    body: { message: 'testing' }
}, clientConfig, rawResponse: false})

console.log('RESPONSE: ', response)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,85 @@ const searchResults = await searchClient.productSearch({
 
 Invalid query parameters that are not a part of the API and do not follow the `c_` custom query parameter convention will be filtered from the request and a warning will be displayed.
 
+### Custom APIs
+
+The SDK supports calling [custom APIs](https://developer.salesforce.com/docs/commerce/commerce-api/guide/custom-apis.html) with a helper function, `customApiHelper.callCustomEndpoint()`.
+
+Example usage:
+
+```javascript
+import * as CommerceSdk from "commerce-sdk";
+const { customApiHelper } = CommerceSdk;
+
+// client configuration parameters
+const clientConfigExample = {
+  parameters: {
+    clientId: "<your-client-id>",
+    organizationId: "<your-org-id>",
+    shortCode: "<your-short-code>",
+    siteId: "<your-site-id>",
+  },
+  // If not provided, it'll use the default production URI:
+  // 'https://{shortCode}.api.commercecloud.salesforce.com/custom/{apiName}/{apiVersion}'
+  // path parameters should be wrapped in curly braces like the default production URI
+  baseUri: "<your-base-uri>"
+};
+
+const access_token = "'<INSERT_ACCESS_TOKEN_HERE>'"
+
+// Required params: apiName, endpointPath, shortCode, organizaitonId
+// Required path params can be passed into:
+// options.customApiPathParameters or clientConfig.parameters
+// customApiPathParameters will take priority for duplicate values
+const customApiArgs = { 
+  apiName: 'loyalty-info',
+  apiVersion: 'v1', // defaults to v1 if not provided
+  endpointPath: 'customers'
+}
+
+const getResponse = await customApiHelper.callCustomEndpoint({ 
+  options: {
+    // http operation is defaulted to 'GET' if not provided
+    method: 'GET',
+    parameters: {
+      queryParameter: 'queryParameter1',
+    },
+    headers: {
+      // Content-Type is defaulted to application/json if not provided
+      'Content-type': 'application/json',
+      authorization: `Bearer ${access_token}`
+    },
+    customApiPathParameters: customApiArgs,
+  }, 
+  clientConfig: clientConfigExample,
+  // Flag to retrieve raw response or data from helper function
+  rawResponse: false
+})
+
+const postResponse = await customApiHelper.callCustomEndpoint({ 
+  options: {
+    method: 'POST',
+    headers: {
+      authorization: `Bearer ${access_token}`
+    },
+    customApiPathParameters: {
+      apiVersion: 'v1',
+      endpointPath: 'greeting',
+      apiName: 'e2e-tests',
+    },
+    // The body will be automatically formatted for 'Content-Type' 
+    // 'application/json' and 'application/x-www-form-urlencoded',
+    // so an object can be passed
+    body: { data: 'data' }
+  }, 
+  clientConfig: clientConfigExample, 
+  rawResponse: false
+})
+
+console.log('get response: ', getResponse)
+console.log('post response: ', postResponse)
+```
+
 ## Caching
 
 The SDK currently supports two types of caches - In-memory and Redis. Both the implementations respect [standard cache headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control). To use another type of cache, write your own implementation of the [CacheManager](https://github.com/SalesforceCommerceCloud/commerce-sdk-core/tree/main/src/base/cacheManager.ts). See the [default cache manager](https://github.com/SalesforceCommerceCloud/commerce-sdk-core/tree/main/src/base/cacheManagerKeyv.ts) to design your implementation.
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -229,6 +229,85 @@ const searchResults = await searchClient.productSearch({
...
 
+### Custom APIs
+
+The SDK supports calling [custom APIs](https://developer.salesforce.com/docs/commerce/commerce-api/guide/custom-apis.html) with a helper function, `customApiHelper.callCustomEndpoint()`.
+
+Example usage:
+
+```javascript
+import * as CommerceSdk from "commerce-sdk";
+const { customApiHelper } = CommerceSdk;
```
bendvc, line 240: Looking at how the other helpers are imported:

`import { Search, Customer, helpers, slasHelpers } from "commerce-sdk";`

Should `customApiHelper` be under the helpers object since it's a singular helper?

bendvc: Maybe `customApiHelper` should be `customApiHelpers` ?

src/static/helpers/customApi.test.ts
```diff
@@ -0,0 +1,262 @@
...
+      ...copyOptions.customApiPathParameters,
+    };
+
+    expect(runFetchSpy.callCount).to.equal(1);
+    expect(runFetchPassedArgs[1].pathParameters).to.deep.equal(
+      expectedPathParams
+    );
+    expect(runFetchPassedArgs[1].client.clientConfig.parameters).to.deep.equal(
+      expectedPathParams
+    );
```
bendvc, line 220: Interesting thing, and I'm making an assumption, is that the function itself accomplishes this "prioritization" by using the spread operator, but we are also using the spread operator to create this "expectedPathParams". So it almost nullifies the test in a way. Maybe we just test directly the values one by one to see if they are the expected values? 

Let me know if I'm wrong in my assumption.

src/static/helpers/customApi.ts
```diff
@@ -0,0 +1,151 @@
...
+ * @param args.clientConfig.baseUri? - baseUri used for the request, where the path parameters are wrapped in curly braces. Default value is 'https://{shortCode}.api.commercecloud.salesforce.com/custom/{apiName}/{apiVersion}'
+ * @param args.clientConfig.fetchOptions? - fetchOptions that are passed onto the fetch request
+ * @param args.clientConfig.throwOnBadResponse? - flag that when set true will throw a response error if the fetch request fails (returns with a status code outside the range of 200-299 or 304 redirect)
+ * @param args.clientConfig.proxy? - Routes API calls through a proxy when set
+ * @param args.rawResponse? - Flag to return the raw response from the fetch call. True for raw response object, false for the data from the response
+ * @returns Raw response or data from response based on rawResponse argument from fetch call
+ */
+export const callCustomEndpoint = async (args: {
```
bendvc, line 47: Dumb question..and maybe this is a comment about helpers in general or static code. Would we not always have code duplication from `commerce-sdk` and `commerce-sdk-isomorphic`. E.g. this is the same code you have in that other lib. Meaning if there is a bug in it, you have to fix it in 2 places, not just one.

README.md
kevinxh: I believe in the isomorphic implementation, the new helper is under the `helpers` object. Do we want to keep it consistent across libs?

This is from the PR description of the isomorphic change.
```js
import {helpers} from 'commerce-sdk-isomorphic';

let response = await helpers.callCustomEndpoint({
    method: 'GET',
    headers: {
        'Content-Type': 'application/json',
        authorization: `Bearer ${access_token}`
    }
}, clientConfig, false)
```

src/static/helpers/customApi.ts
```diff
@@ -0,0 +1,151 @@
...
+    parameters?: {
+      [key: string]: string | number | boolean | string[] | number[];
+    };
+    customApiPathParameters?: {
+      apiName?: string;
+      apiVersion?: string;
+      endpointPath?: string;
+      organizationId?: string;
+      shortCode?: string;
+    };
```
kevinxh, line 59: just double checking here, can you remind me if this is the same interface for the isomorphic helper?

src/static/helpers/customApi.ts
joeluong-sfcc: Yes, this is correct. Unfortunately this is one of the gaps in having two separate SDKs. There were talks in the past for creating a core repo with code that is shared by both SDKs, but it wasn't ever followed up on.

src/static/helpers/customApi.ts
joeluong-sfcc: Yes, this is correct, at least for `customApiPathParameters`. For the entire function, however, there's a new property on the options object `retrySettings` that is available in the node SDK but not the isomorphic SDK

README.md
joeluong-sfcc: Updated to have callCustomEndpoint under `helpers`

src/static/helpers/customApi.ts
```diff
@@ -0,0 +1,167 @@
...
+    headers?: {
+      authorization?: string;
+    } & { [key: string]: string };
+    body?: BodyInit | unknown;
+    retrySettings?: OperationOptions;
+    fetchOptions?: RequestInit;
+    enableTransformBody?: boolean;
```
bendvc, line 80: Aside from the `customApiPathParameters` is this type being used defined anywhere else? Curious if there is an opportunity to reuse an existing type and use typescript magic to make a union with the new addition.

templates/operations.ts.hbs
```diff
@@ -132,7 +132,7 @@
       {{#each ../parameters}}
         "{{{name}}}": optionParams["{{{name}}}"]
         {{! if it is a common path parameter, fallback to client config for undefined }}
-        {{#if (isCommonPathParameter name)}} !== undefined ? optionParams["{{{name}}}"] : configParams["{{{name}}}"]{{/if}}
+        {{#if (isCommonPathParameter name)}} !== undefined ? optionParams["{{{name}}}"] : configParams?.["{{{name}}}"]{{/if}}
```
bendvc, line 135: Why did we require all these changes to use optional chaining in this template?

joeluong-sfcc: I originally had typescript errors from the generated typescript due to configParams potentially being undefined, but I just regenerated the typescript again and no longer see those issues... Might be something wrong with my environment but reverted those changes

src/static/helpers/customApi.ts
joeluong-sfcc: Updated the typing so that `customApiPathParameters` reuses a type instead of defining its own

bendvc (approved):

kevinxh (approved):

bendvc (approved):

PR #415, opened 2025-03-24
W-18089100 - allow custom params in helpers

alexvuong:
This PR allows custom parameters to be passed into `/authorize` endpoint via the `loginGuestUser` and `authorizeIDP` helpers  and `/login` via `loginRegisteredUserB2C` and `loginRegisteredUserB2Cprivate` helper. 

- Customer headers are not allowed and will be ignored via SLAS

Some Note on the authentication endpoints
- The /authorize endpoint does not call SCAPI / OCAPI customers/auth where the hooks take place.
- /login will have authentication hooks. See [here](https://developer.salesforce.com/docs/commerce/commerce-api/guide/hook-method-details.html)

### DIFF
--- a/src/static/helpers/slas.test.ts
+++ b/src/static/helpers/slas.test.ts
@@ -301,6 +301,45 @@ describe("Guest user flow", () => {
     expect(options.body).to.include.keys("code_verifier");
     expect(accessToken).to.be.deep.equals(expectedTokenResponse);
   });
+
+  it("can pass custom params/header when using a public client flow", async () => {
+    const mockSlasClient = createSlasClient();
+    const spy = sinon.spy(mockSlasClient, "getAccessToken");
+    const { shortCode, organizationId } = mockSlasClient.clientConfig
+      .parameters as CommonParameters;
+
+    nock(`https://${shortCode}.api.commercecloud.salesforce.com`)
+      .get(`/shopper/auth/v1/organizations/${organizationId}/oauth2/authorize`)
+      .query(true)
+      .reply(303, { response_body: "response_body" });
+
+    nock(`https://${shortCode}.api.commercecloud.salesforce.com`)
+      .post(`/shopper/auth/v1/organizations/${organizationId}/oauth2/token`)
+      .query(true)
+      .reply(200, expectedTokenResponse);
+
+    const accessToken = await slasHelper.loginGuestUser(
+      mockSlasClient,
+      {
+        redirectURI: parameters.redirectURI,
+        c_cloth: "jeans",
+      },
+      {
+        headers: { c_custom_header: "custom-value" },
+      }
+    );
+
+    const options = spy.getCall(0).args[0];
+
+    // have to match object since code_verifier is randomly generated
+    sinon.assert.match(options, {
+      ...expectedOptionsPublic,
+      parameters: { c_cloth: "jeans" },
+      headers: { c_custom_header: "custom-value" },
+    });
+    expect(options.body).to.include.keys("code_verifier");
+    expect(accessToken).to.be.deep.equals(expectedTokenResponse);
+  });
 });
 
 describe("Registered B2C user flow", () => {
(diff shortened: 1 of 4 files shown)


src/static/helpers/slasClient.ts
```diff
@@ -109,3 +109,19 @@ export interface ISlasClient {
 
   clientConfig: ClientConfig;
 }
+
+/**
+ * Custom query parameter type with any string prefixed with `c_` as the key and the allowed
+ * types for query parameters for the value.
+ */
+export type CustomQueryParameters = {
+  [key in `c_${string}`]: string | number | boolean | string[] | number[];
```
kevinxh, line 118: curious, what's the use case for `string[]` and `number[]`?

src/static/helpers/slasClient.ts
```diff
@@ -109,3 +109,19 @@ export interface ISlasClient {
...
+};
+
+/**
+ * Custom body request type with any string prefixed with `c_` as the key and the allowed
+ * types for the value.
+ */
+export type CustomRequestBody = {
```
kevinxh, line 125: `CustomQueryParameters` and `CustomRequestBody` seems to be the exact same type?

I feel like we had previously added support for `c_` parameters, is there a chance that the type definition is already implemented somewhere in the codebase?

src/static/helpers/slas.test.ts
kevinxh, line 1: Could you add a test case for `c_` parameters / headers?

src/static/helpers/slas.ts
```diff
@@ -92,27 +101,33 @@ export async function authorize(
     redirectURI: string;
     hint?: string;
     usid?: string;
+  } & CustomQueryParameters,
+  options?: {
+    headers?: { [key: string]: string };
```
kevinxh, line 106: I think nodejs does have an official type `IncomingHttpHeaders`, but i haven't used it personally.

src/static/helpers/slas.ts
```diff
@@ -408,23 +480,29 @@ export function refreshAccessTokenPrivate(
...
-      Authorization: `Bearer ${parameters.accessToken}`,
+      Authorization: `Bearer ${accessToken}`,
+      ...options?.headers,
     },
     parameters: {
-      refresh_token: parameters.refreshToken,
+      refresh_token: refreshToken,
       client_id: slasClient.clientConfig.parameters.clientId,
       channel_id: slasClient.clientConfig.parameters.siteId,
+      ...restOfParams,
```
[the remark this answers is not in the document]
alexvuong: parameters here is for query parameters, it won't be translated to request body

src/static/helpers/slas.test.ts
alexvuong: Will do, but it is not necessary to follow c_* patterm. For headers, we are not restricting to just c_string, because APIs accept any custom headers. c_* syntax is for custom query parameters, and for the `autheticateCustomer` request body

src/static/helpers/slasClient.ts
joeluong-sfcc: We're allowing the same query parameter values as defined in the rest of the SDK: https://github.com/SalesforceCommerceCloud/commerce-sdk-isomorphic/blob/da2f6ca1828dafe6625a15751b2c7aedfbf532c9/src/static/helpers/types.ts#L54

`string[]` and `number[]` are valid types for query parameters and we'll want to allow customers to pass those into their API call if they've configured their environment to handle those custom parameters

src/static/helpers/slasClient.ts
joeluong-sfcc: We used type `any` when we originally implemented custom parameters, which should be updated to use the `CustomQueryParameters` in the next major version release, I've documented that work in a ticket as part of the OAS spike

`CustomQueryParameters` and `CustomRequestBody` should not be the same type as you pass in complex objects for the body

In `src/static/helpers/fetchHelper.ts` we use `BodyInit`, it also might make sense here

src/static/helpers/slas.test.ts
alexvuong: Correction, API seems to have a stricter for customer header. Check [here](https://developer.salesforce.com/docs/commerce/commerce-api/guide/extensibility_via_hooks.html). However, I don't think it will throw error if you send a random header without the c_pattern. It may simple ignore that header if you are expecting it to use from the SCAPI hook. Tested locally with a script and I can successfully get back a access_token
```
// GUEST LOGIN
const guestTokenResponse = await helpers
    .loginGuestUser(slasClient, { redirectURI }, {headers: {test: 'hello', c_hello: 'custom'}})
    .then((guestTokenResponse) => {
        console.log("Guest Token Response: ", guestTokenResponse);
        return guestTokenResponse;
    })
    .catch((error) =>
        console.log("Error fetching token for guest login: ", error)
    );
```

src/static/helpers/slas.ts
```diff
@@ -83,6 +90,8 @@ export const generateCodeChallenge = async (
  * @param parameters.redirectURI - the location the client will be returned to after successful login with 3rd party IDP. Must be registered in SLAS.
  * @param parameters.hint? - optional string to hint at a particular IDP. Guest sessions are created by setting this to 'guest'
  * @param parameters.usid? - optional saved SLAS user id to link the new session to a previous session
+ * @param options - an object containing the options for this function.
+ * @param options.headers - optional headers to pass in the ShopperLogin authorizeCustomer method.
```
joeluong-sfcc, line 94: Adding `?` for both `options` and `options.headers?` for consistency in doc comments for here and other functions please

src/static/helpers/slas.ts
```diff
@@ -149,9 +169,10 @@ export async function loginGuestUserPrivate(
     `${slasClient.clientConfig.parameters.clientId}:${credentials.clientSecret}`
   )}`;
 
-  const options = {
+  const opts = {
     headers: {
       Authorization: authorization,
+      ...options?.headers,
```
joeluong-sfcc, line 175: Swap order to ensure we don't override `Authorization`

src/static/helpers/slas.ts
```diff
@@ -308,28 +363,32 @@ export async function loginRegisteredUserB2C(
   const authorization = `Basic ${stringToBase64(
     `${credentials.username}:${credentials.password}`
   )}`;
+  const { usid, redirectURI, ...restOfParams } = parameters;
 
-  const options = {
+  const opts = {
     headers: {
       Authorization: authorization,
+      ...(options?.headers || {}),
```
joeluong-sfcc, line 371: Swap order here as well

src/static/helpers/slas.ts
```diff
@@ -379,26 +450,33 @@ export function refreshAccessToken(
...
 ): Promise<TokenResponse> {
+  const { refreshToken, ...restOfParams } = parameters;
   const authorization = `Basic ${stringToBase64(
     `${slasClient.clientConfig.parameters.clientId}:${credentials.clientSecret}`
   )}`;
-  const options = {
+  const opts = {
     headers: {
       Authorization: authorization,
+      ...options?.headers,
```
joeluong-sfcc, line 470: Swap here as well

joeluong-sfcc (commented):
LGTM besides the minor tweaks and updating `CustomRequestBody` type

kevinxh (approved):

kevinxh (approved):
Good job!

joeluong-sfcc (approved):

PR #419, opened 2025-06-17
@W-18482680: allow unknown query params

unandyala:


joeluong-sfcc (approved):
LGTM, conditional logic makes sense to me, and thanks for adding a test

PR #422, opened 2025-06-24
Fix CI

joeluong-sfcc:


base.tsconfig.json
```diff
@@ -16,6 +16,7 @@
         "noErrorTruncation": true,
         "noImplicitReturns": true,
         "noImplicitThis": true,
-        "preserveConstEnums": true
+        "preserveConstEnums": true,
+        "skipLibCheck": true
```
vcua-mobify, line 20: I'm guessing we're skipping lib check because we removed some types from devDependencies?

vcua-mobify (approved):

base.tsconfig.json
joeluong-sfcc: I updated these files as we didn't need certain changes, but the issue is that we were trying to transpile some typescript in the `node_modules` that was causing an error when you ran `npm run build`

```
> tsc && eslint dist --ext .ts,.js --quiet --fix

node_modules/@rdfjs/types/index.d.ts:3:1 - error TS1383: Only named exports may use 'export type'.

3 export type * from './data-model';
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@rdfjs/types/index.d.ts:4:1 - error TS1383: Only named exports may use 'export type'.

4 export type * from './stream';
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@rdfjs/types/index.d.ts:5:1 - error TS1383: Only named exports may use 'export type'.

5 export type * from './dataset';
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@rdfjs/types/index.d.ts:6:1 - error TS1383: Only named exports may use 'export type'.

6 export type * from './query';
```

alexvuong (approved):

PR #425, opened 2025-07-02
Feat: Remove API Family Layer (@W-18726388@)

joeluong-sfcc:


vcua-mobify (approved):
LGTM

I ran the generator and have verified that the namespaces are no longer there in the generated APIs. APIs like ShopperContext and Catalog which in the past had naming collisions did not have any.

alexvuong (approved):

PR #426, opened 2025-07-16
Feat: Ensure docs site works with OAS (@W-18483003@)

joeluong-sfcc:
This PR updates the docs to reflect the new OAS changes in the Isomorphic SDK. Due to the restructuring of the SDK, there's additional complexity to navigating the docs site. A section has been added in the README for clarity.

This is almost identical to the isomorphic SDK site, however, since there are a large number of APIs, when you navigate to an API class, the right sidebar to navigate through the API classes disappears and you have to click on `Globals` to navigate back. A section was added in the README to describe this difference. 

Synk CI check will be fixed in another PR.

To test this PR:

1. `git checkout ju/oas-docs`
2. run `open docs/index.html`
3. navigate the docs site


package.json
```diff
@@ -80,10 +80,10 @@
     "sort-package-json": "^1.53.1",
     "source-map-support": "^0.5.21",
     "ts-node": "^10.9.2",
-    "typedoc": "^0.17.8",
+    "typedoc": "^0.17.7",
     "typedoc-plugin-external-module-map": "1.2.1",
     "typedoc-plugin-nojekyll": "^1.0.1",
-    "typescript": "~5.1.0"
```
joeluong-sfcc, line 86: Originally upgraded to fix a linting issue, but the linting issue is not appearing any longer so dropping it down to keep it in alignment with commerce-sdk-isomorphic

unandyala:
image
Can we removed quotes around "index", "types", "version"?
Does "types" included common types shared between classes?
What us "version" module? Can we add few more details in the docs?

README.md
adamraya, line 1: Should we update the old Customer namespace pattern example?

https://github.com/SalesforceCommerceCloud/commerce-sdk/blob/bbc5f0ceb7a4ad6c7e555e4ac53b117ecf92e384/README.md#L133

To be consistent with the new pattern:
```
const loginClient = new ShopperLogin(config)
```

joeluong-sfcc: Ah good catch, I forgot to replace that, updated

README.md
```diff
@@ -4,6 +4,27 @@ The Salesforce Commerce SDK allows easy interaction with the Salesforce B2C Comm
...
+
+**Utility Classes:** Utility classes and methods such as `clientConfig` and `helpers` maintain the same structure as previous versions.
+
+**NOTES:** 
+
+1. **Type Access**: API class types are accessible through the `<api_class>Types` namespace (e.g., `ShopperProductsTypes`). Individual types can be accessed as `ShopperProductsTypes.Product`.
+
+2. **Type References**: The `References` section under API classes in the generated documentation may show duplicate entries. This occurs because types are exported both at their original definition and under the API class namespace. Both references point to the same underlying type definition.
+
 ## :warning: Planned API Changes :warning:
```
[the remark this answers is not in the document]
joeluong-sfcc: I added a new point in the documentation section to point to the sample code for an example on how to import using v4 vs v5

PR #432, opened 2025-08-14
Fix snyk CI for feature OAS branch

joeluong-sfcc:
Fixes CI for: https://github.com/SalesforceCommerceCloud/commerce-sdk/pull/423

We had open API generator as a dependency but we don't use it directly, we have it through RAML toolkit so we can remove it, commerce-sdk-isomorphic does not have this dependency either

This PR also bumps up `commerce-sdk-core` version

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,6 @@
       "license": "BSD-3-Clause",
       "dependencies": {
         "@commerce-apps/core": "^1.8.0",
-        "@openapitools/openapi-generator-cli": "^2.20.0",
         "nanoid": "^3.3.4",
         "retry": "^0.13.1",
         "tslib": "^2.4.1"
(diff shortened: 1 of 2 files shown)


vcua-mobify (approved):

package-lock.json
```diff
@@ -2953,6 +2953,7 @@
       "version": "5.0.0",
       "resolved": "https://registry.npmjs.org/@isaacs/brace-expansion/-/brace-expansion-5.0.0.tgz",
       "integrity": "sha512-ZT55BDLV0yv0RBm2czMiZ+SqCGO7AvmOM3G/w2xhVPH+te0aKgFjmBvGlL1dH+ql2tgGO3MVrbb3jCKyvpgnxA==",
+      "dev": true,
```
adamraya, line 2956: Is the `"dev": true,` expected?

PR #447, opened 2026-05-08
Chore: bump version (@W-22400300@)

joeluong-sfcc:


adamraya (dismissed):
Looking at the previous version bump [PR #439](https://github.com/SalesforceCommerceCloud/commerce-sdk/pull/439) I see we also regenerated the docs since `docs/modules/version.html` embeds the SDK version still shows commerce-sdk@6.1.0. Should we regenerate docs as well?

https://github.com/SalesforceCommerceCloud/commerce-sdk/blob/ju/bump_version/docs/modules/version.html#L98

vmarta (dismissed):
Yeah, it looks we need to also update the docs like last time.

adamraya (approved):

vmarta (approved):

PR #452, opened 2026-06-29
Release v6.4.0 for ECOM v26.7 (@W-23232606@)

joeluong-sfcc:
## Summary
- Bumps all SCAPI/OCAPI/SLAS APIs to the versions pinned in the api-manager `release/26.7` POM.
- Bumps SDK version to 6.4.0.

## API version diff vs v6.3.0
- auth-oas-v1 1.48.0 → 1.50.0 (Shopper Login)
- shopper-availability-oas-v1 1.1.0 → 1.2.0
- shopper-baskets-oas-v2 2.5.1 → 2.9.0
- shopper-experience-oas-v1 1.4.1 → 1.5.2
- shopper-orders-oas-v1 1.12.1 → 1.15.0
- shopper-products-oas-v1 1.5.1 → 1.10.1
- shopper-search-oas-v1 1.9.0 → 1.10.0

## CDN Zones API held at 1.2.0
`zones-oas-v1` (`cdn-api-process-apis`) is held at 1.2.0 in this release rather than bumping to 1.3.0. The v1.3.0 spec adds a Turnstile tag, and the OAS generator template at `src/generate-oas.ts:78` returns `CDNZones` as the class name for every operation of the Zones spec regardless of tag — so v1.3.0 produces two files exporting the same `CDNZones` class (`DefaultApi.ts` and `TurnstileApi.ts`) and the TypeScript build fails with `TS2308`. A follow-up will fix the generator to tag-scope the class name; once that lands, a patch can bump CDN Zones to 1.3.0.

## Test plan
- [x] `npm run updateApis` regenerates all API specs cleanly with no `[error]` diff entries that change v6.4.0 semantics.
- [x] `npm run build` (tsc) compiles green.
- [x] `npm run doc` regenerates the version table and HTML docs; v6.3.0 entries are untouched.
- [ ] Reviewer: confirm the held CDN Zones note + W-23232606 are sufficient to ship without v1.3.0.

## Ticket
W-23232606


vcua-mobify (commented):
`zones-oas-v1` verified to be at v 1.2.0

+1

vcua-mobify (approved):

PR #453, opened 2026-07-09
Fix: Collapse CDN Zones spec onto one tag to unblock generation (@W-23234552@)

joeluong-sfcc:
## Summary
- The Zones OAS spec ships a spec-level `x-sdk-classname: CDNZones` that openapi-generator threads to every emitted API file's class. When operations span multiple tags (zones-oas 1.3.0 introduces a `Turnstile` tag), the generator emits one file per tag — each declaring `export class CDNZones` — and the auto-generated barrel fails `TS2308` on the duplicate. Any bump past zones-oas 1.2.0 is blocked.

  Concretely, generating `zones-oas-v1=1.3.0` on `main` (without this fix):

  ```
  renderedTemplates/cdnApiProcessApis/apis/
  ├── DefaultApi.ts     # untagged zone ops    -> export class CDNZones
  ├── TurnstileApi.ts   # Turnstile-tagged ops -> export class CDNZones   <-- duplicate
  └── index.ts          # export * from './DefaultApi'
                        # export * from './TurnstileApi'                   <-- TS2308 here
  ```

  `npm run compile` fails with `TS2308: Module './DefaultApi' has already exported a member named 'CDNZones'` at `apis/index.ts:2`.
- Fix: pass `--openapi-normalizer SET_TAGS_FOR_ALL_OPERATIONS=default` to the openapi-generator invocation for the Zones spec only. That collapses every operation onto the `default` tag, so the generator emits a single `DefaultApi.ts` with all operations (Turnstile included) on one class — preserving the filename convention every other spec in the SDK already uses.

  Generating `zones-oas-v1=1.3.0` on this branch (with the fix):

  ```
  renderedTemplates/cdnApiProcessApis/apis/
  ├── DefaultApi.ts     # every op incl. Turnstile -> export class CDNZones
  └── index.ts          # export * from './DefaultApi'
  ```

  `npm run compile` exits clean.
- Consumer surface is unchanged: `import { CDNZones } from "commerce-sdk"` resolves through the same barrel chain to the same class. At pinned 1.2.0 the emitted `DefaultApi.ts`, model files, and every downstream barrel are byte-identical to base.

## Test plan
- [x] Regression test at `src/generate-oas.test.ts` — proven **red on base** (`TS2305: Module './generate-oas' has no exported member 'resolveGeneratorFlags'`, i.e. the test can't compile without the fix's carrier function) and **green on fix** (52 passing).
- [x] `npm run test:unit` — 52 passing.
- [x] `npm run compile` — clean typecheck.
- [x] `npm run lint` — clean.
- [x] Re-rendered locally at pinned **zones-oas 1.2.0** with and without the fix — `DefaultApi.ts`, `apis/index.ts`, all 98 model files, `models/index.ts`, `cdnApiProcessApis/index.ts`, `cdnApiProcessApis/runtime.ts`, and the top-level `renderedTemplates/index.ts` are byte-identical between base and fix.
- [x] Re-rendered locally at target **zones-oas 1.3.0**: single `DefaultApi.ts`, single `export class CDNZones`, all previous zone endpoints plus the 5 new Turnstile operations (`createTurnstileWidget`, `deleteTurnstileWidget`, `getTurnstileWidget`, etc.) on one class. Clean `apis/index.ts` barrel, no TS2308.

## Reviewer manual smoke test

To reproduce the 1.3.0 generation locally on this branch:

```bash
# 1. Check out the branch
git fetch origin
git checkout ju/cdn-zones-classname-W-23234552
npm install

# 2. Bump the CDN Zones spec pin. Edit api-versions.txt line 5:
#      zones-oas-v1=1.2.0   ->   zones-oas-v1=1.3.0
#    Do NOT commit this edit — it's for local smoke only. The pin bump ships in
#    a separate release ticket.

# 3. Pull the 1.3.0 spec from Anypoint (needs Anypoint creds in your env):
export ANYPOINT_USERNAME=...
export ANYPOINT_PASSWORD=...
npm run updateApis

# 4. Regenerate + build.
npm run renderTemplates
npm run compile
```

Expected result:
- `renderedTemplates/cdnApiProcessApis/apis/` contains exactly one API file, `DefaultApi.ts`, with `export class CDNZones extends BaseClient`. No `TurnstileApi.ts`, no duplicate class.
- `DefaultApi.ts` has all pre-existing zone endpoints plus the 5 new Turnstile operations (`createTurnstileWidget`, `deleteTurnstileWidget`, `getTurnstileWidget`, `getTurnstileWidgets`, `updateTurnstileWidget`).
- 6 new `TurnstileWidget*` model files under `renderedTemplates/cdnApiProcessApis/models/`.
- `npm run compile` exits 0 (no TS2308 on the barrel).

Sanity check on `main` first (optional): repeat steps 2–4 on `main` and `npm run compile` will fail with `TS2308: Module './DefaultApi' has already exported a member named 'CDNZones'` at `renderedTemplates/cdnApiProcessApis/apis/index.ts:2` — this is the failure this PR resolves.

To restore local state after smoke: `git checkout -- api-versions.txt && rm -rf apis/zones-oas-1.3.0 renderedTemplates`.

## Follow-ups
- The `api-versions.txt` bump of `zones-oas-v1` from 1.2.0 to 1.3.0 belongs in a separate release ticket per the ticket's AC.
- `src/generate-oas.ts:58` — `resolveApiName`'s `"CDN API - Process APIs OAS"` branch is dead (no shipping spec uses that name); the current name is `"Zones OAS"`, handled at line 82. Chore ticket to clean up.
- The three parallel per-spec switches (`resolveApiName`, `resolveDirectoryName`, `resolveGeneratorFlags`) could collapse into one data-driven config table. Separate Refactor ticket.
- The mustache template at `templatesOas/apis.endpoint.mustache:242` emits `const pathParams = {}` without a type annotation. VS Code's default `noImplicitAny` lights up every generated `DefaultApi.ts` in the repo with editor squiggles — cosmetic only, the repo `tsconfig.json` compiles cleanly. Chore ticket to add `Record<string, unknown>` to the mustache emit.

## Ticket
W-23234552

### DIFF
--- a/src/generate-oas.test.ts
+++ b/src/generate-oas.test.ts
@@ -9,9 +9,7 @@ import { resolveGeneratorFlags } from "./generate-oas";
 
 type ApiSpecDetail = Parameters<typeof resolveGeneratorFlags>[0];
 
-const buildDetail = (
-  overrides: Partial<ApiSpecDetail> = {}
-): ApiSpecDetail => ({
+const buildDetail = (overrides: Partial<ApiSpecDetail>): ApiSpecDetail => ({
   filepath: "/tmp/spec.yaml",
   filename: "spec.yaml",
   name: "Ignored",
--- a/src/generate-oas.test.ts
+++ b/src/generate-oas.test.ts
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2025, salesforce.com, inc.
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ * For full license text, see the LICENSE file in the repo root or https://opensource.org/licenses/BSD-3-Clause
+ */
+import { expect } from "chai";
+import { resolveGeneratorFlags } from "./generate-oas";
+
+type ApiSpecDetail = Parameters<typeof resolveGeneratorFlags>[0];
+
+const buildDetail = (name: string): ApiSpecDetail => ({
+  filepath: "/tmp/spec.yaml",
+  filename: "spec.yaml",
+  name,
+  apiName: "Ignored",
+  directoryName: "ignored",
+});
+
+describe("resolveGeneratorFlags", () => {
+  it("appends SET_TAGS_FOR_ALL_OPERATIONS for the Zones spec so tagged operations do not split into a second class", () => {
+    const flags = resolveGeneratorFlags(buildDetail("Zones OAS"));
+
+    expect(flags).to.include(
+      "--openapi-normalizer SET_TAGS_FOR_ALL_OPERATIONS=default"
+    );
+    expect(flags).to.include("--reserved-words-mappings delete=delete");
+  });
+
+  it("returns the base flags unchanged for every other spec", () => {
+    const flags = resolveGeneratorFlags(buildDetail("Shopper Baskets OAS"));
+
+    expect(flags).to.equal("--reserved-words-mappings delete=delete");
+  });
+});
--- a/src/generate-oas.test.ts
+++ b/src/generate-oas.test.ts
@@ -9,17 +9,22 @@ import { resolveGeneratorFlags } from "./generate-oas";
 
 type ApiSpecDetail = Parameters<typeof resolveGeneratorFlags>[0];
 
-const buildDetail = (name: string): ApiSpecDetail => ({
+const buildDetail = (
+  overrides: Partial<ApiSpecDetail> = {}
+): ApiSpecDetail => ({
   filepath: "/tmp/spec.yaml",
   filename: "spec.yaml",
-  name,
+  name: "Ignored",
   apiName: "Ignored",
   directoryName: "ignored",
+  ...overrides,
 });
 
 describe("resolveGeneratorFlags", () => {
   it("appends SET_TAGS_FOR_ALL_OPERATIONS for the Zones spec so tagged operations do not split into a second class", () => {
-    const flags = resolveGeneratorFlags(buildDetail("Zones OAS"));
+    const flags = resolveGeneratorFlags(
+      buildDetail({ apiName: "CDNZones", name: "Zones OAS" })
+    );
 
     expect(flags).to.include(
       "--openapi-normalizer SET_TAGS_FOR_ALL_OPERATIONS=default"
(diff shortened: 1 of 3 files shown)


src/generate-oas.ts
```diff
@@ -126,6 +130,20 @@ export function getAPIDetailsFromExchange(directory: string): ApiSpecDetail {
...
 
+/**
+ * The Zones spec ships some operations under a `Turnstile` tag; openapi-generator
+ * would otherwise emit a separate TurnstileApi.ts whose class collides with
+ * DefaultApi.ts on the spec-level `x-sdk-classname: CDNZones`. Collapsing every
+ * operation onto the `default` tag keeps them on one generated class and
+ * preserves the DefaultApi.ts filename shared with every other spec.
+ */
+export function resolveGeneratorFlags(apiSpecDetail: ApiSpecDetail): string {
+  if (apiSpecDetail.name === ZONES_SPEC_NAME) {
```
adamraya: nit: this matches on `name` (`"Zones OAS"`). If Zones ever ships as v2, `name` becomes `"Zones OAS V2"` (see `getAPIDetailsFromExchange`), this check is skipped, and the bug comes back with no failing test.

`apiName` is always `"CDNZones"` no matter the version, so it is safer. Bonus: it also covers the old `"CDN API - Process APIs OAS"` name that `resolveApiName` maps to `CDNZones` too.

```suggestion
  if (apiSpecDetail.apiName === "CDNZones") {
```

src/generate-oas.test.ts
```diff
@@ -0,0 +1,35 @@
...
+  it("appends SET_TAGS_FOR_ALL_OPERATIONS for the Zones spec so tagged operations do not split into a second class", () => {
+    const flags = resolveGeneratorFlags(buildDetail("Zones OAS"));
+
+    expect(flags).to.include(
+      "--openapi-normalizer SET_TAGS_FOR_ALL_OPERATIONS=default"
+    );
+    expect(flags).to.include("--reserved-words-mappings delete=delete");
+  });
+
+  it("returns the base flags unchanged for every other spec", () => {
```
adamraya, line 43: Nice test. If you switch the check to `apiName`, consider adding one more case for a `"Zones OAS V2"` spec so a future v2 rename can't silently bring the bug back.

adamraya (approved):

src/generate-oas.ts
```diff
@@ -137,10 +155,10 @@ export function generateSDKs(apiSpecDetail: ApiSpecDetail): void {
...
       generateFromOas.generateFromOas({
-        inputSpec: `${filepath}`,
+        inputSpec: filepath,
         outputDir: `${outputDir}`,
         templateDir: `${TEMPLATE_DIRECTORY}`,
-        flags: `--reserved-words-mappings delete=delete`,
+        flags: resolveGeneratorFlags(apiSpecDetail),
       });
     } catch (error) {
       console.error(`Error generating SDK for ${name}: ${error}`);
```
adamraya, line 164: nit: this catch logs but does not rethrow, so `renderTemplates` still exits 0 when a spec fails to generate. A bad normalizer flag would fail silently and only surface later at compile.

Rethrow so the build fails loudly:

```suggestion
      console.error(`Error generating SDK for ${name}: ${error}`);
      throw error;
```

src/generate-oas.ts
joeluong-sfcc: Good catch, done — `resolveGeneratorFlags` now keys off `apiSpecDetail.apiName === "CDNZones"`, so a future Zones v2 rename and the legacy `"CDN API - Process APIs OAS"` title are both covered.

src/generate-oas.test.ts
joeluong-sfcc: Done — added a `"Zones OAS V2"` case in `src/generate-oas.test.ts` that asserts the normalizer flag still gets appended when only the spec title changes.

src/generate-oas.ts
joeluong-sfcc: Good callout, done — `generateSDKs` now rethrows after the console.error so a bad flag fails the render immediately instead of surfacing as a downstream compile error.

